PR #1, opened 2018-02-09
First functional POC installing Ginkgo LMS & CMS with Docker

sampaccoud:
## Purpose

The purpose was to have functional LMS and CMS containers running a pure Docker approach without Ansible. The full project can be started with a simple `make bootstrap` command that:
- clones repositories
- builds docker images
- updates assets
- migrates database
- imports a demo course
- creates a super user (user input required in console)

## Loophole

Docker's best practices could not be applied completely because of the way Open edX manages depencies (e.g. "pip install -e ."). Build time on code updates with this PR is less than 5 minutes, not too bad. But we could make it nearly immediate with some refactoring to the way Open edX handles dependencies.

If you want to go further, build the project with `make all`. When it's done, cd into `src/edx-platform` and modify the `.gitignore` file to keep only the `*.pyc` rule. Now run `git status` to see all the files that have been modified as a result of running `pip install requirements/local.txt` (last step of the Docker file).

These modifications to the repository:
- force us to reinstall the last step when mounting the source files from host (which is required for development in order to see our changes immediately). Not hard but hackish and unexpected,
- forces Docker to reinstall the last step on each build. We could speed up build time to make it almost instantaneous.
- is a bit obscure for the new comer because it is not a common pattern. Dependencies and project code are usually kept separate. 

Maybe improving this is not so hard and the benefits could be big?

For reference, here is the output of the `git status` command after removing gitignore:
```
    .prereqs_cache/
    Open_edX.egg-info/
    cms/envs/fun_platform.py
    common/lib/calc/calc.egg-info/
    common/lib/capa/capa.egg-info/
    common/lib/chem/chem.egg-info/
    common/lib/dogstats/dogstats_wrapper.egg-info/
    common/lib/safe_lxml/safe_lxml.egg-info/
    common/lib/sandbox-packages/sandbox_packages.egg-info/
    common/lib/symmath/symmath.egg-info/
    common/lib/xmodule/XModule.egg-info/
    common/static/bundles/
    common/static/common/js/vendor/
    lms/envs/fun_platform.py
    lms/static/certificates/css/
    lms/static/css/bootstrap/
    lms/static/css/discussion/
    lms/static/css/lms-course-rtl.css
    lms/static/css/lms-course.css
    lms/static/css/lms-footer-edx-rtl.css
    lms/static/css/lms-footer-edx.css
    lms/static/css/lms-footer-rtl.css
    lms/static/css/lms-footer.css
    lms/static/css/lms-learner-dashboard-rtl.css
    lms/static/css/lms-learner-dashboard.css
    lms/static/css/lms-main-v1-rtl.css
    lms/static/css/lms-main-v1.css
    lms/static/css/lms-main-v2-rtl.css
    lms/static/css/lms-main-v2.css
    node_modules/
    openedx/core/lib/xblock_builtin/xblock_discussion/xblock_discussion.egg-info/
    themes/red-theme/lms/static/css/bootstrap/
    themes/red-theme/lms/static/css/discussion/
    themes/red-theme/lms/static/css/lms-course-rtl.css
    themes/red-theme/lms/static/css/lms-course.css
    themes/red-theme/lms/static/css/lms-footer-edx-rtl.css
    themes/red-theme/lms/static/css/lms-footer-edx.css
    themes/red-theme/lms/static/css/lms-footer-rtl.css
    themes/red-theme/lms/static/css/lms-footer.css
    themes/red-theme/lms/static/css/lms-learner-dashboard-rtl.css
    themes/red-theme/lms/static/css/lms-learner-dashboard.css
    themes/red-theme/lms/static/css/lms-main-v1-rtl.css
    themes/red-theme/lms/static/css/lms-main-v1.css
    themes/red-theme/lms/static/css/lms-main-v2-rtl.css
    themes/red-theme/lms/static/css/lms-main-v2.css
    themes/stanford-style/lms/static/css/bootstrap/
    themes/stanford-style/lms/static/css/discussion/
    themes/stanford-style/lms/static/css/lms-course-rtl.css
    themes/stanford-style/lms/static/css/lms-course.css
    themes/stanford-style/lms/static/css/lms-footer-edx-rtl.css
    themes/stanford-style/lms/static/css/lms-footer-edx.css
    themes/stanford-style/lms/static/css/lms-footer-rtl.css
    themes/stanford-style/lms/static/css/lms-footer.css
    themes/stanford-style/lms/static/css/lms-learner-dashboard-rtl.css
    themes/stanford-style/lms/static/css/lms-learner-dashboard.css
    themes/stanford-style/lms/static/css/lms-main-v1-rtl.css
    themes/stanford-style/lms/static/css/lms-main-v1.css
    themes/stanford-style/lms/static/css/lms-main-v2-rtl.css
    themes/stanford-style/lms/static/css/lms-main-v2.css
```


settings/cms.env.json
```diff
@@ -0,0 +1,69 @@
+{
+  "SITE_NAME": "localhost:8001",
+  "BOOK_URL": "",
+  "LOG_DIR": "/data/logs",
+  "LOGGING_ENV": "sandbox",
+  "OAUTH_OIDC_ISSUER": "http://localhost:8000/oauth2",
+  "PLATFORM_NAME": "FUN Platform Studio - Open edX",
+  "FEATURES": {
+    "PREVIEW_LMS_BASE": "localhost:8000"
```
regisb: I don't think this is going to work, because port 8000 is mapped on port 8072. When clicking on "preview" the user will be redirected to "localhost:8000" which will not succeed because of port mapping. I'm afraid the only elegant solution is to not change ports on docker port mapping. i.e: 

for the lms:

```
ports:
        - "8000:8000"
```

for the cms:

```
ports:
        - "8001:8001"
```

You will also need to patch the cms command:

```
python manage.py cms --settings=fun_platform runserver 0.0.0.0:8001
```

settings/fun_platform_cms.py
```diff
@@ -0,0 +1,23 @@
+
```
jmaupetit: Empty first line is not required.

settings/fun_platform_lms.py
```diff
@@ -0,0 +1,22 @@
+
```
jmaupetit: Empty first line is not required.

settings/fun_platform_cms.py
```diff
@@ -0,0 +1,23 @@
...
+FEATURES['ENABLE_DISCUSSION_SERVICE'] = False
+
+ALLOWED_HOSTS = ["*"]
+
+LOGGING = get_logger_config(LOG_DIR,
+                            logging_env=ENV_TOKENS['LOGGING_ENV'],
+                            debug=False,
+                            dev_env=True,
+                            service_variant=SERVICE_VARIANT)
```
jmaupetit: Fix indentation issue 😛 

```python
LOGGING = get_logger_config(
    LOG_DIR,
    logging_env=ENV_TOKENS['LOGGING_ENV'],
    debug=False,
    dev_env=True,
    service_variant=SERVICE_VARIANT
)
```

settings/fun_platform_lms.py
```diff
@@ -0,0 +1,22 @@
...
+FEATURES['ENABLE_DISCUSSION_SERVICE'] = False
+
+ALLOWED_HOSTS = ["*"]
+
+LOGGING = get_logger_config(LOG_DIR,
+                            logging_env=ENV_TOKENS['LOGGING_ENV'],
+                            debug=False,
+                            dev_env=True,
+                            service_variant=SERVICE_VARIANT)
```
jmaupetit: Can you fix indentation here?

settings/fun_platform_cms.py
```diff
@@ -0,0 +1,23 @@
...
+
+LOGGING['handlers']['local'] = LOGGING['handlers']['tracking'] = {
+    'class': 'logging.NullHandler',
+}
```
jmaupetit: IMO this pattern is hard to read, i would prefer:

```python
LOGGING['handler'].update(
    local={'class': 'logging.NullHandler'},
    tracking={'class': 'logging.NullHandler'}
}
```

settings/fun_platform_lms.py
```diff
@@ -0,0 +1,22 @@
...
+
+LOGGING['handlers']['local'] = LOGGING['handlers']['tracking'] = {
+    'class': 'logging.NullHandler',
+}
```
jmaupetit: Same remark about the multiple assignment pattern.

docker-compose.yml
```diff
@@ -0,0 +1,85 @@
...
+    command: mongod --smallfiles --nojournal --storageEngine wiredTiger
+
+  memcached:
+    image: memcached:1.4
+
+  rabbitmq:
+    image: rabbitmq:3.6
+
+  lms:
+    image: fun-platform-edxapp
```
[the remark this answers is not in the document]
sampaccoud: so I replace `image` by `build`

settings/fun_platform_cms.py
sampaccoud: removed

settings/fun_platform_cms.py
sampaccoud: hao de

settings/fun_platform_cms.py
sampaccoud: agreed

settings/fun_platform_lms.py
sampaccoud: removed

settings/fun_platform_lms.py
sampaccoud: yes I can

settings/fun_platform_lms.py
sampaccoud: same answer

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Install system dependencies
+RUN apt-get update && apt-get upgrade -y && \
+    apt-get install -y build-essential curl g++ gcc gettext gfortran git git-core \
+    graphviz graphviz-dev language-pack-en libffi-dev libfreetype6-dev libgeos-dev \
+    libjpeg8-dev liblapack-dev libmysqlclient-dev libpng12-dev libxml2-dev \
+    libxmlsec1-dev libxslt1-dev nodejs nodejs-legacy npm ntp pkg-config python-apt python-dev \
+    python-pip software-properties-common swig
+# Remove the package lists after installation is a good practice
+# RUN rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
jmaupetit: I think the idea is to reduce the number of layers and their size. If you split `apt` instructions and this clean up command, the size of the built image will get bigger for (apparently) no reason.

Can you clarify why do you want to condition this command by the context (production vs dev)?

Makefile
```diff
@@ -0,0 +1,70 @@
...
+# Django
+MANAGE_CMS       = $(COMPOSE_RUN_CMS) python manage.py cms --settings=fun_platform
+MANAGE_LMS       = $(COMPOSE_RUN_LMS) python manage.py lms --settings=fun_platform
+
+all: bootstrap run
+
+default: help
+
+bootstrap:  ## install development dependencies
+	@./bin/clone_repositories.sh
```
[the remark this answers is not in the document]
jmaupetit: My bad, it was. 👍

README.md
```diff
@@ -0,0 +1,47 @@
...
+
+- **mysql:** the SQL database used to structure and persist the application data,
+- **mongodb:** the no-SQL database used to store course content,
+- **memcached:** the cache engine,
+- **rabbitmq:** the messaging broker used for asynchronous tasks,
+- **lms:** the Django web application used by learners,
+- **cms:** the Django web application used by teachers,
+- **forum:** the Ruby web application serving the discussion forum **TODO**
+- **ecommerce:** The Django web application used for payments **TODO**
+- **xqueue:** the interface for the LMS to communicate with external grader services **TODO**
```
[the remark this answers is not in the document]
jmaupetit: Well spotted! 😉

Dockerfile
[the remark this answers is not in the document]
sampaccoud: Because in dev you might want to install additional packages to test something and you will have to manually reactivate sources?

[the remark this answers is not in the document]
jmaupetit: 🤔 If you want to install packages for test or development purpose, why not doing something like:

```
$ docker-compose exec lms bash
root@ddfb978bacbb:/app/edx-platform# apt-get update
root@ddfb978bacbb:/app/edx-platform# apt-get install foo
```

And when you succeed/conclude, update the package list in your `Dockerfile`?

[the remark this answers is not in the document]
sampaccoud: legit. I thought more manipulation was necessary to reactivate the package list. I have modified my Docker file.

Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM ubuntu:16.04
+
+# Install system dependencies
+# Removing the package lists after installation is a good practice
+RUN apt-get update && apt-get upgrade -y && \
+    apt-get install -y build-essential curl g++ gcc gettext gfortran git git-core \
+    graphviz graphviz-dev language-pack-en libffi-dev libfreetype6-dev libgeos-dev \
+    libjpeg8-dev liblapack-dev libmysqlclient-dev libpng12-dev libxml2-dev \
+    libxmlsec1-dev libxslt1-dev nodejs nodejs-legacy npm ntp pkg-config python-apt python-dev \
+    python-pip software-properties-common swig \
```
jmaupetit: the `&&` are missing at the end of the `install` command

Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM ubuntu:16.04
+
+# Install system dependencies
+# Removing the package lists after installation is a good practice
+RUN apt-get update && apt-get upgrade -y && \
```
jmaupetit: Can you separate those two commands in two lines for readability?

settings/cms.env.json
sampaccoud: Not sure I understand what's the purpose of preview yet :-/ 
Or should I declare?
PREVIEW_LMS_BASE : "localhost:8072"

bin/clone_repositories
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+echo -e "\033[1;33mGetting edx-platform\033[0m"
+mkdir -p src/edx-platform
+# Clone it from here to get a better message in command line
+git clone https://github.com/edx/edx-platform.git src/edx-platform
+cd src/edx-platform
+git checkout open-release/ginkgo.master
```
smarnach: This branch is not guaranteed to be always in a working state.  I recommend sticking to the actual tagged releases, like `open-release/ginkgo.1`.

Also note that `git clone` will implicitly create the target directory, and that you can use `-b ref` to check out a specific branch or tag.  So the whole contents of this script is equivalent to the single command

    git clone -b open-relase/ginkgo.1 https://github.com/edx/edx-platform.git src/edx-platform

docker-compose.yml
```diff
@@ -0,0 +1,85 @@
...
+  rabbitmq:
+    image: rabbitmq:3.6
+
+  lms:
+    build: .
+    image: fun-platform-edxapp
+    ports:
+        - "8072:8000"
+    volumes:
+      - ./src/edx-platform:/app/edx-platform
```
smarnach: I don't fully understand the details here.  The `Dockerfile` copies the repo into the image, but it seems to be also mounted into the image as a volume in the same location.  Is this intentional?  What's the idea behind it?  Or did I misunderstand this somehow?

docker/etc/nginx/conf.d/cms.conf
```diff
@@ -0,0 +1,42 @@
...
+
+server {
+  listen 8081;
+  server_name localhost;
+  
+  # Prevent invalid display courseware in IE 10+ with high privacy settings
+  add_header P3P 'CP="Open edX does not have a P3P policy."';
+
+  # Nginx does not support nested condition or conditions so
+  # there is an unfortunate mix of conditonals here.
```
smarnach: It looks like this comment was left here, though it doesn't make sense in this context.

docker/etc/nginx/conf.d/cms.conf
```diff
@@ -0,0 +1,42 @@
+upstream cms-backend {
```
smarnach: Do we really need nginx for the development setup?  In production it's required, but for development we usually don't need it.

settings/cms.env.json
smarnach: The preview domain is used for new content in Studio that hasn't been published yet.  If you are in a unit with unpublished changes in Studio, there is a preview link somewhere (I think in the top right corner), and when you click it you see a preview of the content.  This feature isn't really that important in a dev environment, but should be configured correctly in production.

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+WORKDIR /app/edx-platform
+
+# Install Python requirements
+# ... adding only targeted requirements files first to benefit from caching
+ADD ./src/edx-platform/requirements/edx /app/edx-platform/requirements/edx
+RUN pip install --src ../src -r requirements/edx/pre.txt
+RUN pip install --src ../src -r requirements/edx/github.txt
+RUN pip install --src ../src -r requirements/edx/base.txt
+RUN pip install --src ../src -r requirements/edx/paver.txt
+RUN pip install --src ../src -r requirements/edx/post.txt
```
smarnach: Since this seems to be for development, you probably also want to install the requirements from `development.txt` and `testing.txt`, so you can run the tests.

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Install Javascript requirements
+# ... adding only the package.json file first to benefit from caching
+ADD ./src/edx-platform/package.json /app/edx-platform/package.json
+RUN npm install
```
smarnach: The node packages are installed in a subdirectory of `edx-platform`.  Since the docker-compose file below seems to mount `edx-platform` as a volume on top of the directory in the Docker image, I'm not sure whether the packages will still be visible when the container is run.

Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM ubuntu:16.04
+
+# Install system dependencies
+# Removing the package lists after installation is a good practice
```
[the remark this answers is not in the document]
sampaccoud: I read this:
```
Note: The official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required."
```
Here:
https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/#avoid-installing-unnecessary-packages

Makefile
```diff
@@ -0,0 +1,67 @@
...
+default: help
+
+bootstrap:  ## install development dependencies
+	@./bin/clone_repositories;
+	${MAKE} build;
+	${MAKE} run;
+	${MAKE} update-assets;
+	${MAKE} migrate;
+	${MAKE} demo-course;
+	${MAKE} superuser;
```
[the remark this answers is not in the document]
sampaccoud: The idea was to preserve the "make help" command.

Makefile
```diff
@@ -0,0 +1,67 @@
...
+	@$(COMPOSE) build;
+	# Reinstall the local requirements with volumes activated to allow
+	# mounting the project sources from host for development.
+	@$(COMPOSE_RUN_LMS) pip install --src ../src -r requirements/edx/local.txt;
+.PHONY: build
+
+demo-course:  ## import demo course from edX repository
+	@./bin/clone_demo_course
+	docker-compose run --rm -v $(shell pwd)/src/edx-demo-course:/app/edx-demo-course cms \
+	python manage.py cms --settings=fun_platform import /data/media /app/edx-demo-course
```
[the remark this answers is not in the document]
sampaccoud: I could not because of the -v option. Any idea?

bin/clone_repositories
sampaccoud: Nice!

docker-compose.yml
sampaccoud: Yes the idea is to build a working image that (with a bit more work) will go to production. Docker compose is then used for development and mounting volumes from the host so that we can code and see our changes immediately.

sampaccoud: I'm going to explain this in documentation since it was confusing.

docker/etc/nginx/conf.d/cms.conf
sampaccoud: Yes. The nginx conf must be adapted for production, we have a lot of microservices that will be served through this container. For this "first POC" I just copy/pasted what I understood from edx/configuration.

docker/etc/nginx/conf.d/cms.conf
sampaccoud: Yes the idea is that fun-platform will go to production. Gunicorn will be the CMD in the dockerfile and runserver only the docker-compose override of CMD.

sampaccoud: I will try to make it work in this PR for clarity.

settings/cms.env.json
sampaccoud: I don't understand what this setting does. It's just sending the teacher to the lms so why would we need to declare a different domain?

Dockerfile
sampaccoud: Maybe I do this with a context so that the production image doesn't get it.

Dockerfile
sampaccoud: Well spotted. I will run "npm install" again from docker-compose so that it is also available on the devstack.

bin/clone_repositories
sampaccoud: In fact not so nice because when the repository already already exists, this will break and do nothing whereas my script was still checking out the desired branch. So I kept the script and added a comment to explain why it's better.

Dockerfile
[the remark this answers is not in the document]
smarnach: Ah, thanks, good to know!

bin/clone_repositories
smarnach: OK, it wasn't clear to me what the desired behaviour is then.

settings/cms.env.json
smarnach: @sampaccoud Because it will show changes that haven't been published yet.  I didn't look into the details why the used a different domain name, but I presume it's to be able to use the same 
URL path/code/whatever as for the main LMS site.  It would certainly be possible to use a different path on the usual LMS domain, but edX decided to go with a different domain instead (which seems like an ok choice to make to me).

docker/etc/nginx/conf.d/cms.conf
sampaccoud: @smarnach done. We now have separate images for production and development.

README.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Getting started
+
+Make sure you have a recent version of [Docker](https://docs.docker.com/install) and [Docker Compose](https://docs.docker.com/compose/install) installed on your laptop:
+
+    $ docker -v
+      Docker version 17.12.0-ce, build c97c6d6
+
+    $ docker-compose --version
+      docker-compose version 1.17.1, build 6d101fb
```
jmaupetit: Maybe we can add a note about buggy docker-compose 1.19.0 release?

README.md
```diff
@@ -0,0 +1,141 @@
...
+    * served by nginx: [localhost:8083](http://localhost:8083)
+
+See other available commands by running:
+
+    $ make
+
+
+## TODO list
+
+- Investigate why the CMD does not received system signals,
```
jmaupetit: Which system signals are you spotting?

README.md
```diff
@@ -0,0 +1,141 @@
...
+- Add missing services (ecommerce, forum, xqueue),
+- Run `paver update_assets --skip-collect` from within the Dockerfile and run `collectstatic` only after deployment to production,
+- Make use of Docker multi-stage builds to remove build tools from the production Docker image (node_modules, nodejs, git, python-pip, etc.),
+- Improve Python dependencies on [edx-platform](https://github.com/edx/edx-platform).
+
+### About this last point
+
+Docker's best practices could not be applied completely because of the way edx-platform manages depencies. Build time on code updates with this PR is a few minutes depending on your connection, not too bad. But we could make it nearly immediate with some refactoring to the way edx-platform handles dependencies.
```
jmaupetit: `s/depencies/dependencies/`

bin/clone_demo_course
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+echo -e "\033[1;33mGetting edx-demo-course\033[0m"
+# Clone it from here to get a better message in command line
+git clone https://github.com/edx/edx-demo-course.git src/edx-demo-course
+cd src/edx-demo-course
+# Better do the checkout separately from the clone so that if the repository
+# already exists, the checkout is still done to the branch we want.
+git checkout open-release/ginkgo.1
+cd -
```
jmaupetit: This is not required, unless you plan to add other actions to this script.

bin/clone_repositories
```diff
@@ -0,0 +1,12 @@
...
+echo -e "\033[1;33mGetting edx-platform\033[0m"
+# Clone it from here to get a better message in command line
+git clone https://github.com/edx/edx-platform.git src/edx-platform
+cd src/edx-platform
+# Clean node installations in case the repo pre-existed
+git clean -fdx
+# Better do the checkout separately from the clone so that if the repository
+# already exists, the checkout is still done to the branch we want.
+git checkout open-release/ginkgo.1
+cd -
```
jmaupetit: This is not required, unless you plan to add other actions to this script.

README.md
sampaccoud: "docker stop" sends a SIGTERM signal and our application does not handle it. It ends up being killed after 10 seconds which is annoying and is dangerous. We need to fix it. 
As for nginx, the proper signal to send for a graceful shutdown seems to be a SIGQUIT so we need to wrap stop to change the signal sent?

As you see it's not so straightforward so I left it as a TODO.

README.md
sampaccoud: done

README.md
sampaccoud: done

PR #2, opened 2018-02-28
Add default development configurations and improve container tags

jmaupetit:
## Container tag schema

We will stick to Open edX release tags, e.g. `ginkgo.1`, for `fun-platform-edxapp` container tags.

A development container tag will be suffixed by `-dev`, e.g. `ginkgo.1-dev`

## Default configurations for development

This PR adds default configurations to containers used for development purpose. This is useful to avoid having to copy all configuration files when its not required (_i.e._ if not overridden).

### DIFF
--- a/Dockerfile_dev
+++ b/Dockerfile_dev
@@ -9,3 +9,6 @@ RUN apt-get update && \
 
 RUN pip install --src ../src -r requirements/edx/testing.txt && \
     pip install --src ../src -r requirements/edx/development.txt
+
+# Copy default configurations for development
+COPY ./config /config/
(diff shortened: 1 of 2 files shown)


Dockerfile_dev
```diff
@@ -9,3 +9,6 @@ RUN apt-get update && \
 
 RUN pip install --src ../src -r requirements/edx/testing.txt && \
     pip install --src ../src -r requirements/edx/development.txt
+
+# Copy default configurations for development
+COPY ./config/* /config/
```
sampaccoud: I think you don't need the * ?

jmaupetit: True. Fixed.

PR #3, opened 2018-03-01
ci(circle) add build & build-dev jobs

jmaupetit:
NB: The build-dev job depends on the build job to speed up build time by using production container cache.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,262 @@
+# fun-platform
+# Docker containers CI
+version: 2
+
+# List fun-platform jobs that will be integrated and executed in a workflow
+jobs:
+
+  # Build job
+  # Build the Docker container ready for production
+  build:
+
+    # We use the machine executor, i.e. a VM, not a container
+    machine:
+      # Cache docker layers so that we strongly speed up this job execution
+      docker_layer_caching: true
+
+    working_directory: ~/fun
+
+    steps:
+      # Checkout fun-platform sources
+      - checkout
+
+      # Restore the ~/fun/src cached repository. If the cache does not exists for
+      # the current .Revision (commit hash), we fall back to the latest cache
+      # with a label matching 'edx-v1-'
+      - restore_cache:
+          keys:
+            - edx-v1-{{ .Revision }}
+            - edx-v1-
+
+      # Clone Open edX sources
+      - run:
+          name: Clone Open edX platform
+          command: |
+            bin/clone_repositories
+
+      # Production container build, it will be tagged as edxapp:latest
+      - run:
+          name: Build container
+          command: |
+            docker build -t edxapp:latest .
+
+      # Cache Open edX repository (cloned in ~/fun/src) as the checkout is
+      # rather time consuming for this project
+      - save_cache:
+          paths:
+            - ~/fun/src
+          key: edx-v1-{{ .Revision }}
+
+      # Save and compress production container to ~/docker/edxapp.tgz
+      - run:
+          name: Save container
+          command: |
+            mkdir ~/docker
+            docker save edxapp | gzip > ~/docker/edxapp.tgz
+
+      # Cache containers directory to load them in another job
+      - save_cache:
+          paths:
+            - ~/docker
+          key: docker-v1-{{ .Revision }}
+
+  # Build dev job
+  # Build the Docker container ready for development
+  build-dev:
+
+    machine:
+      docker_layer_caching: true
+
+    working_directory: ~/fun
+
+    steps:
+      - checkout
+
+      - restore_cache:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,157 @@
...
+          key: edx-v1-{{ .Revision }}
+      - run:
+          name: Save container
+          command: |
+            mkdir ~/docker
+            docker save edxapp | gzip > ~/docker/edxapp.tgz
+      - save_cache:
+          paths:
+            - ~/docker
+          key: docker-v1-{{ .Revision }}
```
sampaccoud: Didn't get the logic in the name of each step?

.circleci/config.yml
```diff
@@ -0,0 +1,157 @@
...
+
+  hub:
+    machine: true
+    steps:
+      - restore_cache:
+          keys:
+            - docker-v1-{{ .Revision }}
+      - restore_cache:
+          keys:
+            - docker-dev-v1-{{ .Revision }}
```
sampaccoud: why do you need the dev cache for the production image ?

.circleci/config.yml
```diff
@@ -0,0 +1,157 @@
...
+            - build-dev
+          filters:
+            tags:
+              only: /.*/
+      - test:
+          requires:
+            - build-dev
+          filters:
+            tags:
+              only: /.*/
```
sampaccoud: ok, I would need much more comments to understand this file. Maybe someone who knows circleCI can review.

.circleci/config.yml
```diff
@@ -0,0 +1,158 @@
...
+            bin/clone_repositories
+      - run:
+          name: Build container
+          command: |
+            docker build -t edxapp:latest .
+      - run:
+          name: Tag container
+          command: |
+            docker tag edxapp fundocker/edxapp:ginkgo.1
+            docker images edxapp
```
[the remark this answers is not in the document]
jmaupetit: It stands for logging purpose, I want to check tags during the build.

.circleci/config.yml
jmaupetit: Do you mean step names or cache keys?

.circleci/config.yml
jmaupetit: I need to load both production and development images to tag them and push them to docker hub.

.circleci/config.yml
jmaupetit: What can I do to improve this?

.circleci/config.yml
sampaccoud: I meant step names. The fact that there are duplicates.
I think I would find it nice that each step has a comment line (I mean Ansible style where the name of each task is long enough to explain what it does)

.circleci/config.yml
sampaccoud: I couldn't think of a reason to push the dev image should go to docker hub. WDYT ?

.circleci/config.yml
sampaccoud: see above. One comment per step ?

.circleci/config.yml
jmaupetit: CircleCI configuration file format does not allow custom names for built-in steps such as `save_cache`. But as suggested below, I can add comments.

.circleci/config.yml
jmaupetit: This is the image you'll need to develop an Open edX application such as Fonzie (with default configurations and installed dependencies).

.circleci/config.yml
jmaupetit: Good idea.

.circleci/config.yml
```diff
@@ -0,0 +1,262 @@
...
+      # This script requires CI environment variable to run; those are passed to
+      # the container via the `-e` docker run option:
+      #
+      # - CIRCLECI: used by the CI script to select the CI strategy (depending
+      #   on the CI platform)
+      # - CIRCLE_NODE_INDEX/CIRCLE_NODE_TOTAL: used for the parallel execution
+      #   of the test suite
+      # - EDXAPP_TEST_MONGO_HOST: the mongo db host used in tests
+      #
+      # We also need to mount the git repository that is used to inegrate
```
sampaccoud: s/inegrate/integrate

.circleci/config.yml
```diff
@@ -0,0 +1,262 @@
...
+
+  # We have a single workflow
+  edxapp:
+
+    jobs:
+
+      # The build job has no required jobs, hence this will be our first job
+      - build:
+          # Filtering rule to run this job: none (we accept all tags; this job
+          # will always run).
```
sampaccoud: :1st_place_medal: +1:

PR #12, opened 2018-03-19
ci(circle) restrict hub job to the master branch

jmaupetit:
We do not want to push PR builds to DockerHub, this fix improves the situation by restricting the push job to the master branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -196,8 +196,8 @@ workflows:
           requires:
             - build-dev
           filters:
-            tags:
-              only: /.*/
+            branches:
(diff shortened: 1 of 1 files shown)


sampaccoud:
Or should it be only push tags of a certain format + commit is on master branch?

jmaupetit:
> Or should it be only push tags of a certain format

Yes we can. But we are not prepared for that yet. For now, it urges to prevent PR's container uploads!

> check that commit is on master branch?

We cannot handle this in CircleCI workflow filters. It needs to be done in a script executed as a job step.

PR #13, opened 2018-03-29
Build and push Docker images with tags matching their git tag

sampaccoud:
We are pushing to Docker only images that are the result of a tag respecting the pattern:
    **{branch_name}-{semver}**

Where branch name is composed of two words representing:
    1) edx-platform full version name,
    2) FUN's specific fork project.

Valid examples:
    * dogwood.3-funmooc-1.0.0
    * eucalyptus-funwb-2.3.19-rc3

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,11 +22,11 @@ jobs:
 
       # Restore the ~/fun/src cached repository. If the cache does not exists for
       # the current .Revision (commit hash), we fall back to the latest cache
-      # with a label matching 'edx-v1-'
+      # with a label matching 'edx-repository-v1-'
       - restore_cache:
           keys:
-            - edx-v1-{{ .Revision }}
-            - edx-v1-
+            - edx-repository-v1-{{ .Revision }}
+            - edx-repository-v1-
 
       # Clone Open edX sources
       - run:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -38,13 +38,13 @@ jobs:
...
-            docker build -t edxapp:latest .
+            docker build -t edxapp:${CIRCLE_TAG} .
 
       # Cache Open edX repository (cloned in ~/fun/src) as the checkout is
       # rather time consuming for this project
       - save_cache:
           paths:
-            - ~/fun/src
+            - ~/fun/src/edx-platform/.git
           key: edx-v1-{{ .Revision }}
```
jmaupetit: You'll need to invalidate the cache by increasing the version number in the prefix, _i.e._ `edx-v2-{{ .Revision }}`. And do not forget to also restore this new cache (in the `restore_cache/keys` section).

.circleci/config.yml
```diff
@@ -192,12 +201,23 @@ workflows:
...
+      #
+      # Regex to validate Semver was taken from: https://github.com/semver/semver/issues/232
       - hub:
           requires:
             - build-dev
           filters:
-            branches:
-              only: master
+            tags:
+              only: /^[a-z0-9.]*-[a-z]*-(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?$/
```
jmaupetit: Why not hard-coding allowed edx releases like:

```
^(dogwood|eucalyptus)...
```

sampaccoud: I thought we  might want to add new forks lol. 
Touch wood we don't need to... but maybe we will add branches on which we want to test something ? and there will also be new versions of edx-platform that will come out etc. So I decided to be lenient on this so that it is not a pain. 

Note that I'm checking in the "hub" job that the branch name is included in the tag name. So we are sure that you won't be able to make a typo in the tag name.

.circleci/config.yml
sampaccoud: I thought maybe this is not necessary because we are doing: 
git checkout -f -b eucalyptus-funwb openfun/fun/whitebrand

so it will fetch the branch from the remote and update the cache anyway. The cache is in fact just a way to download less but we have no risk to end-up with the wrong this. 

I will make a PR on master to add this behavior as well?

jmaupetit: You're right, as we fall back on the latest revision, the cache kind of auto-invalidates at each new commit.

.circleci/config.yml
jmaupetit: > So I decided to be lenient on this so that it is not a pain.

Okay.

> So we are sure that you won't be able to make a typo in the tag name.

Unless you copy/paste the same typo :wink:

PR #16, opened 2018-04-05
Allow Docker container to handle timezone DST correctly

rmoch:
https://serverfault.com/a/856593


Dockerfile
```diff
@@ -1,5 +1,15 @@
 FROM ubuntu:16.04
 
+# Set container timezone and related timezones database and DST rules
+# See https://serverfault.com/a/856593
+ENV TZ 'Europe/Paris'
```
jmaupetit: I think we should default to UTC.

Dockerfile
```diff
@@ -1,5 +1,15 @@
...
+RUN echo $TZ > /etc/timezone && \
+    apt-get update && apt-get install -y tzdata && \
```
jmaupetit: I think you should add `tzdata` to the dependencies listed in the `apt` block below and configure the timezone after this `apt` block.

sampaccoud: maybe this should be done in the same RUN as the one below as it is also doing "apt-get update" that takes time ?

PR #18, opened 2018-04-20
Refactor settings for improved configuration management

sampaccoud:
This PR proposes a refactor of the settings with the objective to build on the `common.py` file instead of the `aws.py` file because:

* going down one level makes things easier to follow,
* the `aws.py` file was strongly targetted to hosting on... AWS,
* allow customizing settings via `yaml` files instead of `json` files:
    + can be generated by a configuration management tool like Ansible,
    + accept comments.
* include in the Docker image all the setting files necessary to start the LMS and CMS with a default configuration. This default configuration should not require any `yaml` file. Configuration can then be fully customized with `yaml` files mounted in the container.
* be able to commit the differences between environments (feature/staging/preprod/production). **e.g.** environments other than `production` should send emails to a mail catcher, only the `development` environment should have DEBUG set to true, etc. We don't want to leave this up to configuration management !

Most of the code is in 2 huge `docker_run_production.py` files. The way I generated them is:

- copy/paste the `aws.py` file,
- replace AUTH_TOKENS from `json` files by a `config` helper loaded from `yaml` files,
- remove all the things that were obviously targetting AWS,
- improve things when I could,
- adapt all the things that could not work in a Docker container,
- shake it until the app starts,
- leave all the rest as it was in `aws.py`. We should continue improving the configuration after deploying it to a real test instance in `OpenShift`.


config/lms/utils.py
```diff
@@ -0,0 +1,66 @@
...
+        else:
+            settings = settings or {}
+
+        # Load the content of a credentials.vault.yaml file placed in the current directory if any
+        # This file is were sensitive settings are stored for a given environment
+        try:
+            with open(os.path.join(dir, 'credentials.vault.yaml')) as f:
+                credentials = yaml.load(f.read())
+        except IOError:
+            print "IOError"
```
jmaupetit: print fail :grinning:

PR #29, opened 2018-06-19
Fix course import containing accentuated filenames

### LINKED ISSUE #28: Course import is not working
## Expected Behavior
We should be able to import a complete course as a tar.gz in the CMS 

## Actual Behavior
When doing so, the import never go further "Updating course" step and stay there forever

capture d ecran 2018-06-19 a 09 30 15

## Steps to Reproduce
go to fun-mooc.fr to export an actual course, CMS: tools>export
in your fun-platform instance CMS: Create new course then: tools>import

## Specifications
dogwood branches

rmoch:
fix #28

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,9 @@ RUN apt-get update && \
     python-pip ruby software-properties-common swig tzdata && \
     rm -rf /var/lib/apt/lists/*
 
+ENV LC_ALL C.UTF-8
+ENV LANG C.UTF-8
+
 # Set container timezone and related timezones database and DST rules
 # See https://serverfault.com/a/856593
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -11,6 +11,9 @@ RUN apt-get update && \
     python-pip ruby software-properties-common swig tzdata && \
     rm -rf /var/lib/apt/lists/*
 
+ENV LC_ALL C.UTF-8
+ENV LANG C.UTF-8
+
```
sampaccoud: Actually, this commit should be a PR to master and then you will rebase the dogwood branch onto master.

The other commit in this PR only concerns dogwood so ok to make it a PR on this branch. 

WDYT?

PR #34, opened 2018-08-01
Make Celery workers great again

jmaupetit:
## Purpose

After having tested recent work related to celery configuration, we needed to make minor adjustments to make celery workers functional.

## Proposal

- [x] Add `redis` python dependency
- [x] Configure Celery to only accept json-serialized messages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ RUN echo $TZ > /etc/timezone && \
 WORKDIR /edx/app/edxapp/edx-platform
 
 # Install Python requirements
+# -- edx requirements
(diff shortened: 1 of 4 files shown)


requirements/fun/base.txt
```diff
@@ -0,0 +1,2 @@
+# Base FUN requirements
```
sampaccoud: I wouldn't call it FUN. It seems that it is something specific to us when in fact it is necessary to run edx with Redis as is now the preferred solution... Call it "specific.txt" ? I'm ok with anything else but FUN ;-)

PR #35, opened 2018-08-02
Configure Celery to only accept json-serialized messages

jmaupetit:
## Purpose

After having tested recent work related to celery configuration, we needed to make minor adjustments to make celery workers functional when running in a container context (using OpenShift).

## Proposal

- [x] format configs with `black`
- [x] Configure Celery to only accept json-serialized messages

### DIFF
--- a/config/cms/docker_run_production.py
+++ b/config/cms/docker_run_production.py
@@ -8,7 +8,9 @@ import platform
 from lms.envs.fun.utils import Configuration
 from openedx.core.lib.derived import derive_settings
 from path import Path as path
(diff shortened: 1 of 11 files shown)


config/cms/docker_run_production.py
```diff
@@ -70,6 +70,12 @@
 
 CELERY_ROUTES = "cms.celery.Router"
 
+# Force accepted content to "json" only. If we also accept pickle-serialized
+# messages, the worker will crash when it's running with a privileged user (even
+# if it's not the root user but a user belonging to the root group, which is our
```
sampaccoud: which is the case with OpenShift for example :smile_cat:

PR #36, opened 2018-08-27
Make theme configurable in docker build

sampaccoud:
## Purpose

We want to be able to set a custom theme when building the Docker image in an OpenShift buildconfig. Since the config volumes are not mounted in buildconfigs, we propose to allow setting it via an environment variable.

## Proposal

- [x] allow setting config parameters via environment variables
- [x] make the "DEFAULT_SITE_THEME" setting configurable in build settings.

Note: I also moved the docker_build.py file to /lms and /cms so that it works the same as in Hawthorn and we can do the same thing in Arnold to deploy...

### DIFF
--- a/config/cms/docker_build_production.py
+++ b/config/cms/docker_build_production.py
@@ -1,0 +1,27 @@
+# This is a minimal settings file allowing us to run "update_assets"
+# in the Dockerfile for the production image of the edxapp CMS
+
+from path import Path as path
+
+from lms.envs.fun.utils import Configuration
+
+from ..common import *
+
+
+# Load custom configuration parameters
+config = Configuration()
+
+# This is a minimal settings file allowing us to run "update_assets"
+# in the Dockerfile
+
+DATABASES = {"default": {}}
+
+XQUEUE_INTERFACE = {"url": None, "django_auth": None}
+
+# We need to override STATIC_ROOT because for CMS, edX appends the value of
+# "EDX_PLATFORM_REVISION" to it by default and we don't want to use this.
+# We should use Django's ManifestStaticFilesStorage for this purpose.
+STATIC_ROOT = path('/edx/app/edxapp/staticfiles')
+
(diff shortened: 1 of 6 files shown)


config/cms/docker_build_production.py
```diff
@@ -0,0 +1,27 @@
...
+
+XQUEUE_INTERFACE = {"url": None, "django_auth": None}
+
+# We need to override STATIC_ROOT because for CMS, edX appends the value of
+# "EDX_PLATFORM_REVISION" to it by default and we don't want to use this.
+# We should use Django's ManifestStaticFilesStorage for this purpose.
+STATIC_ROOT = path('/edx/app/edxapp/staticfiles')
+
+# Allow setting a custom theme
+DEFAULT_SITE_THEME = config("DEFAULT_SITE_THEME", default=None)
```
rmoch: I think we must add to this file 2 other constants: 
ENABLE_COMPREHENSIVE_THEMING sould be set to True to activate theming
COMPREHENSIVE_THEME_DIRS a list of folders containing the themes

sampaccoud: Yes it's true. I didn't add them because the default values are what we want so we don't need to set them for the build... 

For the run, they are configurable thanks to what is in docker_run_production.py

config/lms/utils.py
```diff
@@ -64,9 +64,13 @@ def __call__(self, var_name, *args, **kwargs):
...
-                "or  secrets.yml` file.".format(var_name)
-            )
+            try:
+                return os.environ[var_name]
+            except KeyError:
+                if 'default' in kwargs:
+                    return kwargs['default']
+
+        raise ImproperlyConfigured(
+            'Please set the "{:s}" variable in a `settings.yml` file, a secrets.yml file '
```
jmaupetit: Can you remove back quotes to the settings.yml file name?

PR #37, opened 2018-08-30
♻️(build) make theme configurable in docker build

sampaccoud:
## Purpose

Port https://github.com/openfun/fun-platform/pull/36 to Hawthorn 

## Proposal

I re-applied the modifications manually to the master branch where Hawthron lives.

### DIFF
--- a/config/cms/docker_build_production.py
+++ b/config/cms/docker_build_production.py
@@ -3,9 +3,13 @@
 
 from openedx.core.lib.derived import derive_settings
 
+from lms.envs.fun.utils import Configuration
 from path import Path as path
 
 from ..common import *
+
+# Load custom configuration parameters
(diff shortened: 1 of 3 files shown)


config/cms/docker_build_production.py
```diff
@@ -17,6 +21,9 @@
 STATIC_URL = "/static/studio/"
 STATIC_ROOT = path("/edx/app/edxapp/staticfiles/studio")
 
+# Allow setting a custom theme
+DEFAULT_SITE_THEME = config("DEFAULT_SITE_THEME", default=None)
```
sampaccoud: @jmaupetit I will remove this later if you confirm that it is not necessary because update_assets builds all themes anyway...

PR #38, opened 2018-08-30
🐛(statics) remove webpack stats from STATIC_ROOT

jmaupetit:
# Purpose

There is no obvious reason to store `webpack-stats.json` files in the `STATIC_ROOT` tree. It causes more problem than it solves as we need to manually copy it during deployment since it is not collected via collectstatics.

## Proposal

- [x] Update target file path in production settings for both LMS and CMS


sampaccoud:
@jmaupetit Note that you need to port this PR to the hawthorn.1 branch as well. This is now the master branch that will follow edx-platform's master :laughing:

PR #43, opened 2018-09-08
Make emails work in non production environments

sampaccoud:
## Purpose

In non-production environments, we don't want  to send real emails but we want to easily consult the emails sent to anyone.

## Proposal

The solution for this is to use a mail catcher. There are several solutions available and we chose https://github.com/sj26/mailcatcher because it is easy, good enough and distributed as a Docker image.

Quite a lot of refactoring was necessary to make the most sense of this functionality and allow running it in [Arnold](https://github.com/openfun/arnold):

- [x] make email settings smarter and more flexible with good defaults for each environment,
- [x] allow formatting settings passed as environment variables, :open_mouth: 
- [x] move environment variables definitions to env files,
- [x] disable Celery in fun-platform (while [Arnold](https://github.com/openfun/arnold) running the development environment still uses it),
- [x] add a mailcatcher container and send all emails to it. :tada:

Not related:
- [x] fix "update_assets" in the development Dockerfile


config/lms/utils.py
```diff
@@ -65,13 +67,13 @@ def __call__(self, var_name, *args, **kwargs):
             return self.settings[var_name]
         except KeyError:
             try:
-                return os.environ[var_name]
+                return formatter(os.environ[var_name])
```
sampaccoud: @jmaupetit this will make you happy! you can now do:
```
DATABASES = config("DATABASES", default={...}, formatter=json.loads)
```
and set you environment variable with a json object. Is it :tada: ?

docker-compose.yml
```diff
@@ -27,6 +27,11 @@ services:
   rabbitmq:
     image: rabbitmq:3.6
 
+  mailcatcher:
+    image: sj26/mailcatcher:latest
```
jmaupetit: I would have preferred a tagged release, but only `latest` seems to get pushed to dockerhub :disappointed:

jmaupetit:
Commit message d5483d3d141a0422ac22369deb73cdc8771d9749 seems truncated.

docker-compose.yml
sampaccoud: Yeah, that's why I took "latest". We could easily add a Dockerfile to https://github.com/openfun/openshift-docker if you think this is a problem. Note that it is only running in dev.

PR #45, opened 2018-09-11
Make all settings configurable by environment variables

sampaccoud:
## Purpose

Some settings expect other types of information than strings (boolean, dictionary, list, date, integer, etc.). When passing settings via environment variables we must specify a formatter to convert the
string to the expected type. 

## Proposal

This PR just adds formatters to all the settings.

### DIFF
--- a/config/lms/docker_run_production.py
+++ b/config/lms/docker_run_production.py
@@ -80,6 +80,7 @@ CELERY_QUEUES = config(
         LOW_PRIORITY_QUEUE: {},
         HIGH_MEM_QUEUE: {},
     },
+    formatter=json.loads,
 )
 
 CELERY_ROUTES = "lms.celery.Router"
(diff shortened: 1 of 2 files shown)


config/lms/docker_run_production.py
```diff
@@ -1064,16 +1176,20 @@
...
+    default=STUDENTMODULEHISTORYEXTENDED_OFFSET,
+    formatter=int,
 )
 
 # Cutoff date for granting audit certificates
-if config("AUDIT_CERT_CUTOFF_DATE", default=None):
-    AUDIT_CERT_CUTOFF_DATE = dateutil.parser.parse(config("AUDIT_CERT_CUTOFF_DATE"))
+AUDIT_CERT_CUTOFF_DATE = config(
+    "AUDIT_CERT_CUTOFF_DATE", default=None, formatter=dateutil.parser.parse
+)
```
jmaupetit: :+1:

PR #49, opened 2018-09-12
🐛(dev) add vim editor to dev container

rmoch:
## Purpose
Add vim text editor to development container


sampaccoud:
@rmoch note that you also need to port it to the hawthorn.1 as the master branch is deploying edx-platform's master branch, not hawthorn.1.

PR #52, opened 2018-09-13
Improve flexibility of database related settings

sampaccoud:
## Purpose

We want to be able to configure separately each part of the database related settings: database name versus password for example.

This is because they can come from different sources and be fed to the application through different means. For example:

- the password comes from a vault and is injected via environment variables
- the database name comes from a yaml file and is injected by mounting the file in the container,
- the database host is updated at each deployment and injected via an environment variable.

## Proposal

I refactored the database related settings so that:
- the image works in our docker-compose environment with its default settings
- it is possible to override independently each setting: host, port, user, db name, password, etc.
- it is possible to override the whole configuration (for example to add a read-only replica for heavy read load...). 

While working on this PR, I encountered 2 minor bugs which I fixed:

- Remove translatability of default platform name and description. It breaks because of a bug in Open edX that can only be fixed via a PR to https://github.com/openfun/edx-platform
- Update the demo course to hawthorn.1


bin/clone_demo_course
```diff
@@ -6,4 +6,6 @@ git clone https://github.com/edx/edx-demo-course.git src/edx-demo-course
 cd src/edx-demo-course
 # Better do the checkout separately from the clone so that if the repository
 # already exists, the checkout is still done to the branch we want.
-git checkout master
+git fetch --tags edx open-release/hawthorn.1
+git checkout -f open-release/hawthorn.1
+git clean -fdx
```
jmaupetit: A carriage return is missing here.

bin/clone_demo_course
```diff
@@ -6,4 +6,6 @@ git clone https://github.com/edx/edx-demo-course.git src/edx-demo-course
 cd src/edx-demo-course
 # Better do the checkout separately from the clone so that if the repository
 # already exists, the checkout is still done to the branch we want.
-git checkout master
+git fetch --tags edx open-release/hawthorn.1
```
jmaupetit: :1st_place_medal:

config/cms/docker_run_production.py
```diff
@@ -317,8 +317,11 @@
         "level": "ERROR",
     }
 
-PLATFORM_NAME = config("PLATFORM_NAME", default=PLATFORM_NAME)
-PLATFORM_DESCRIPTION = config("PLATFORM_DESCRIPTION", default=PLATFORM_DESCRIPTION)
+# FIXME: the PLATFORM_NAME and PLATFORM_DESCRIPTION settings should be set to lazy translatable
+# strings but edX tries to serialize them with a default json serializer which breaks. We should
+# submit a PR to fix it in edx-platform
+PLATFORM_NAME = config("PLATFORM_NAME", default="Your platform name here")
+PLATFORM_DESCRIPTION = config("PLATFORM_DESCRIPTION", default="Your platform description here")
```
jmaupetit: I think it might break Arnold's edxapp bootstrap test. But there is no big deal. We'll fix that later.

config/cms/docker_run_production.py
```diff
@@ -394,16 +397,25 @@
...
 USER_TASKS_ARTIFACT_STORAGE = COURSE_IMPORT_EXPORT_STORAGE
 
+# Databases
+
+DATABASE_ENGINE = config("DATABASE_ENGINE", default="django.db.backends.mysql")
+DATABASE_HOST = config("DATABASE_HOST", default="mysql")
+DATABASE_PORT = config("DATABASE_PORT", default=3306, formatter=int)
+DATABASE_NAME = config("DATABASE_NAME", default="edxapp")
+DATABASE_USER = config("DATABASE_USER", default="edxapp_user")
+DATABASE_PASSWORD = config("DATABASE_PASSWORD", default="password")
```
jmaupetit: :+1: for having extracted those variable names. This improves readability over a big `DATABASES` block with embedded defaults.

PR #54, opened 2018-09-18
Smooth UX for development

jmaupetit:
## Purpose

fun-platform development workflow was broken and needed a small refactoring.

## Proposal

We've separated the `dev` from `run` target to make them complementary. If you intend to run production images, use `make run`, while if you want to develop on the platform itself, you can use the `make dev` target.

For development, we now mount `edx-platform` sources and our configurations to the `(l|c)ms-dev` service container. As those are local directories "unprepared" to run `edxapp`, we must replay the `update_assets` command and recreate symlinks between our two volumes.


Makefile
```diff
@@ -30,12 +30,27 @@ demo-course:  ## import demo course from edX repository
...
+# symlinks that points to our custom configuration
+dev:  ## start the cms and lms services (development image and servers)
+	$(COMPOSE_RUN) --no-deps lms-dev \
+		rm /edx/app/edxapp/edx-platform/lms/envs/fun;
+	$(COMPOSE_RUN) --no-deps lms-dev \
+		rm /edx/app/edxapp/edx-platform/cms/envs/fun;
+	$(COMPOSE_RUN) --no-deps lms-dev \
+		ln -sf /config/lms /edx/app/edxapp/edx-platform/lms/envs/fun;
+	$(COMPOSE_RUN) --no-deps cms-dev \
+		ln -sf /config/cms /edx/app/edxapp/edx-platform/cms/envs/fun;
```
jmaupetit: Not sure this is optimal, but it's required to avoid symlink inceptions

Makefile
```diff
@@ -30,12 +30,27 @@ demo-course:  ## import demo course from edX repository
...
+		paver update_assets --settings=fun.docker_build_development --skip-collect;
+.PHONY: update_assets
+
+# As we mount edx-platform as a volume in development, we need to re-create
+# symlinks that points to our custom configuration
+dev:  ## start the cms and lms services (development image and servers)
+	$(COMPOSE_RUN) --no-deps lms-dev \
+		rm /edx/app/edxapp/edx-platform/lms/envs/fun;
+	$(COMPOSE_RUN) --no-deps lms-dev \
+		rm /edx/app/edxapp/edx-platform/cms/envs/fun;
```
jmaupetit: Not sure this is optimal, but it's required to avoid symlink inceptions

Dockerfile_dev
```diff
@@ -9,6 +9,3 @@ RUN apt-get update && \
 
 RUN pip install --src ../src -r requirements/edx/testing.txt && \
     pip install --src ../src -r requirements/edx/development.txt
-
-# Update assets skipping collectstatic (not required for development)
-RUN NO_PREREQ_INSTALL=1 paver update_assets --settings=fun.docker_build_development --skip-collect
```
sampaccoud: I thought of it, if you remove this, we will nee to run it upon deployment in Arnold?

Makefile
```diff
@@ -30,12 +30,24 @@ demo-course:  ## import demo course from edX repository
...
-	@$(COMPOSE) up -d
-	docker cp $(shell docker-compose ps -q lms-dev):/edx/app/edxapp/edx-platform src/
-	@$(COMPOSE) -f docker-compose.yml -f dev-volumes.yml up -d
+# As we mount a local statics directory and edx-platform sources in development,
+# we need to regenerate some required static files and install dependencies in
+# these local directories. Hence, you will need to run the update_assets target
+# everytime you update edx-platform sources and plan to develop in it.
+update_assets:  ## run update_assets to copy required statics in local volumes
+	$(COMPOSE_RUN) lms-dev \
+		paver update_assets --settings=fun.docker_build_development --skip-collect;
```
sampaccoud: so this will uninstall everything and reinstall. As you included it in bootstrap it will take forever even though most people will not use it :-(  Should the bootstrap dev be something separate?

Makefile
```diff
@@ -30,12 +30,24 @@ demo-course:  ## import demo course from edX repository
...
 
-dev:  ## activate source overrides for development
-	# Mount the edx-platform repository volume and synchronize the host with
-	# the container so that we can see our changes immediately.
-	@$(COMPOSE) up -d
-	docker cp $(shell docker-compose ps -q lms-dev):/edx/app/edxapp/edx-platform src/
-	@$(COMPOSE) -f docker-compose.yml -f dev-volumes.yml up -d
+# As we mount a local statics directory and edx-platform sources in development,
+# we need to regenerate some required static files and install dependencies in
+# these local directories. Hence, you will need to run the update_assets target
```
sampaccoud: "local directories" you mean "directories on the host". Not clear here but it's key to understanding that you are not generating these files in the container and it's ok that it dies after the run.

docker-compose.yml
```diff
@@ -53,11 +53,16 @@ services:
...
       - ./config:/config
     command: >
       python manage.py lms runserver 0.0.0.0:8000 --settings=fun.docker_run_development
     depends_on:
-      - lms
+      - mailcatcher
+      - mysql
+      - mongodb
+      - memcached
+      - rabbitmq
```
sampaccoud: Take this opportunity to remove rabbitmq because we now use CELERY_ALWAYS_EAGER = True in fun-platform and anyway we have switched to Redis for Celery...

README.md
```diff
@@ -1,33 +1,39 @@
...
+flexible and fully operable by developpers.
 
+**We hope this kind of `Docker` configuration can soon be included in the
+edx-platform repository itself.**
 
 ## Approach
 
-This project builds docker images that are ready for production. `Docker compose` is only used for development purposes so that we can code locally and see our changes immediately:
+This project builds docker images that are ready for production. `Docker compose` is only used for development purposes so that we can code locally and
+see our changes immediately:
```
sampaccoud: weird wrapping?

README.md
```diff
@@ -1,33 +1,39 @@
...
+Docker compose also allows us to run a complete project in development,
+including database services which in production are not run on Docker. See the
+[docker-compose file](./docker-compose.yml) for details on each service:
 
-- **mysql:** the SQL database used to structure and persist the application data,
+- **mysql:** the SQL database used to structure and persist the application
+  data,
 - **mongodb:** the no-SQL database used to store course content,
 - **memcached:** the cache engine,
 - **rabbitmq:** the messaging broker used for asynchronous tasks,
```
sampaccoud: to be removed

README.md
```diff
@@ -39,66 +45,124 @@ Make sure you have a recent version of [Docker](https://docs.docker.com/install)
...
+- CMS served by runserver at: [http://localhost:8082](http://localhost:8082)
 
+### Hacking with themes
+
+To work on a particular theme, we invite you to use the `paver watch_assets`
+command; _e.g._:
+
+```bash
+$ docker-compose exec lms-dev \
+    paver watch_assets --settings=fun.docker_build_development
```
sampaccoud: surprised that you didn't make it a make target!?

README.md
```diff
@@ -39,66 +45,124 @@ Make sure you have a recent version of [Docker](https://docs.docker.com/install)
...
+
+```conf
+# /etc/sysctl.conf (debian based)
+fs.inotify.max_user_watches=524288
+```
 
 # Docker images
 
-The plan is to prepare several lines of images, the docker files and settings of which are living in their own branch (e.g. ginkgo.1, eucalyptus-funwb, dogwood-funmooc,...)
+The plan is to prepare several lines of images, the docker files and settings of
```
sampaccoud: s/lines/flavors?

README.md
```diff
@@ -39,66 +45,124 @@ Make sure you have a recent version of [Docker](https://docs.docker.com/install)
...
 ## TODO list
 
 - Investigate why the CMD does not received system signals,
 - Add missing services (ecommerce, forum, xqueue),
-- Make use of Docker multi-stage builds to remove build tools from the production Docker image (node_modules, nodejs, git, python-pip, etc.),
-- Improve Python dependencies on [edx-platform](https://github.com/edx/edx-platform).
+- Make use of Docker multi-stage builds to remove build tools from the
+  production Docker image (node_modules, nodejs, git, python-pip, etc.),
+- Improve Python dependencies on
+  [edx-platform](https://github.com/edx/edx-platform).
```
sampaccoud: I think this TODO list should be removed as we will not be able to do these optimizations with the state of edx-platform.

README.md
```diff
@@ -39,66 +45,124 @@ Make sure you have a recent version of [Docker](https://docs.docker.com/install)
...
 
 ### About this last point
```
sampaccoud: should we remove this chapter as well?

sampaccoud:
While you're at it, with @rmoch we thought of adding a line in the production Dockerfile to copy the webpack.json file to /common/static so that we can remove the hack in Arnold.

Dockerfile_dev
jmaupetit: I'm not sure it's required as we already did it on the base production image. Unless development dependencies add new apps... Anyway, this would only be required for the `development` `env_type`?

Makefile
jmaupetit: You are right: I should remove `update_assets` from the `bootstrap` target dependencies.

IMO running `update_assets` once at every new `edx-platform` checkout is not that painful and way more explicit / less hazardous than copying files from a running container.

docker-compose.yml
jmaupetit: I've switched to a `redis` service in case we want to test things in fun-platform.

README.md
jmaupetit: updated :wink:

README.md
jmaupetit: True!

README.md
jmaupetit: Yes.

Dockerfile_dev
sampaccoud: well I don't know exactly how it works but the build is completely different for dev and is given to a different webpack configuration. No pipeline enabled either.

Makefile
sampaccoud: agreed

Makefile
sampaccoud: I meant that someone who has little understanding of these mechanics, will understand if by local you mean "on the host" or "in the container". So maybe:

> In development, we work with local directories (on our host machine) for static files and for edx-platform sources, and mount them in the container (using Docker volumes).

README.md
```diff
@@ -1,33 +1,40 @@
...
+including database services which in production are not run on Docker. See the
+[docker-compose file](./docker-compose.yml) for details on each service:
 
-- **mysql:** the SQL database used to structure and persist the application data,
+- **mysql:** the SQL database used to structure and persist the application
+  data,
 - **mongodb:** the no-SQL database used to store course content,
 - **memcached:** the cache engine,
-- **rabbitmq:** the messaging broker used for asynchronous tasks,
+- **redis:** the messaging broker used for asynchronous tasks,
```
sampaccoud: sorry my comment was confusing: there will not be any redis either in fun-platform. We bypass celery with the setting CELERY_ALWAYS_EAGER=True. You may mention that this is necessary in production but not in this list of docker-compose services?

Makefile
```diff
@@ -50,6 +50,10 @@ dev:  ## start the cms and lms services (development image and servers)
 	$(COMPOSE) up -d cms-dev
 .PHONY: dev
 
+watch_assets:  ## start assets watcher (front-end development)
```
sampaccoud: I think we use "-" in our make targets on other projects

README.md
jmaupetit: Okay, I've removed it.

Dockerfile_dev
sampaccoud: @jmaupetit should we have 2 separate dockerfiles?
- one only used by the CI to build a complete development image for deployment,
- one for real development in docker-compose

README.md
```diff
@@ -39,66 +45,105 @@ Make sure you have a recent version of [Docker](https://docs.docker.com/install)
...
+
+## Developer guide
+
+If you intend to work on edx-platform or its configuration, you'll first need to
+compile and collect static files in local directories that are mounted as docker
+volumes in the target container:
+
+```bash
+$ make update-assets
+$ make collectstatics
```
sampaccoud: "collectstatic" is not required for the dev environment because runserver is finding the files directly from the sources.

But you should keep collectstatic in the make bootstrap so that nginx gets the files?

Dockerfile_dev
jmaupetit: I have no strong feelings about it, but I think the current configuration does not require having two separated docker-compose definitions. It's pretty much explicit.

docker-compose.yml
```diff
@@ -51,13 +49,15 @@ services:
...
       - "8072:8000"
     volumes:
-      - ./data/static:/edx/app/edxapp/staticfiles
-      - ./data/media:/edx/var/edxapp/media
+      - ./src/edx-platform:/edx/app/edxapp/edx-platform
       - ./config:/config
     command: >
       python manage.py lms runserver 0.0.0.0:8000 --settings=fun.docker_run_development
     depends_on:
-      - lms
```
sampaccoud: I had an error because of this change. The lms-dev does not depend on lms anymore so he will try to start before the lms but the build is only defined on the lms.
lms-dev assumes that the image is already there...

jmaupetit: I understand. It's not that easy to handle another way than manually, _i.e_:

```bash
$ docker-compose build lms
$ docker-compose build lms-dev 
```

Because there is one thing for sure: I don't want to start the `lms` every time I start the `lms-dev` service...

README.md
```diff
@@ -39,66 +45,104 @@ Make sure you have a recent version of [Docker](https://docs.docker.com/install)
...
+```
+
+**Tip:** you will need to update assets at every new `edx-platform` checkout.
+
+Now you can start services development server _via_:
+
+```bash
+$ make dev
+```
+
```
rmoch: I think `make dev` must be done before `make update-assets`
if not settings are not yet mounted in `lms-dev` container

jmaupetit: Don't forget that the production image already have a copy of our settings, hence, if `update-assets` does not depend on recent changes (made since the production image build), you don't need to.

sampaccoud: @jmaupetit yes you still need to because the symbolic links are dead and make dev needs to re-write them.

jmaupetit: My bad. You are right. In this case, instead of running `make dev` before `make update-assets`, create a new `create-symlinks` target that is a requirement for both `dev` and `update-assets`:

```Makefile
create-symlinks:
	$(COMPOSE_RUN) --no-deps lms-dev bash -c "\
		rm -f /edx/app/edxapp/edx-platform/lms/envs/fun && \
		rm -f /edx/app/edxapp/edx-platform/cms/envs/fun && \
		ln -sf /config/lms /edx/app/edxapp/edx-platform/lms/envs/fun && \
		ln -sf /config/cms /edx/app/edxapp/edx-platform/cms/envs/fun"
.PHONY: create-symlinks
		
update-assets: create-symlinks
	# [...]
.PHONY: update-assets

dev: create-symlinks
	$(COMPOSE) up -d cms-dev
.PHONY: dev
```

WDYT?

sampaccoud: Sure! I thought it's ok to first run "make dev" as a prerequisite to running "make dev-assets" but you're right let's allow any order.

.circleci/config.yml
```diff
@@ -29,6 +29,14 @@ jobs:
             - edx-repository-v1-{{ .Revision }}
             - edx-repository-v1-
 
+      # Install a recent docker-compose release
```
sampaccoud: you need to update the README

Dockerfile
```diff
@@ -1,51 +1,195 @@
...
-# Install system dependencies
-# Removing the package lists after installation is a good practice
+# Change release to build, by providing the EDXAPP_RELEASE build argument to
+# your build command:
+#
+# $ docker build \
+#     --build-arg EDXAPP_RELEASE="open-release/hawthorn.1" \
+#     -t edxapp:hawthorn.1 \
+#     .
+ARG EDXAPP_RELEASE=release-2018-08-29-14.14
```
jmaupetit: Maybe it should default to master?

bin/clone_repositories
```diff
@@ -1,15 +1,20 @@
 #!/usr/bin/env bash
 
-echo -e "\033[1;33mGetting edx-platform\033[0m"
+RELEASE="${1:-release-2018-08-29-14.14}"
```
jmaupetit: Maybe it should also default to `master`?

Dockerfile
jmaupetit: Ok then.

Dockerfile
```diff
@@ -1,16 +1,179 @@
...
-    python-pip software-properties-common swig tzdata && \
+    apt-get install -y \
+    dnsutils \
+    iputils-ping \
+    libsqlite3-dev \
+    mongodb \
+    vim && \
     rm -rf /var/lib/apt/lists/*
 
+# Install python dependencies
```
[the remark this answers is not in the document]
jmaupetit: Not really outdated but I can improve it.

Makefile
```diff
@@ -1,64 +1,101 @@
...
-	# Mount the edx-platform repository volume and synchronize the host with
-	# the container so that we can see our changes immediately.
-	@$(COMPOSE) up -d
-	docker cp $(shell docker-compose ps -q lms-dev):/edx/app/edxapp/edx-platform src/
-	@$(COMPOSE) -f docker-compose.yml -f dev-volumes.yml up -d
+# In development, we work with local directories (on our host machine) for
+# static files and for edx-platform sources, and mount them in the container
+# (using Docker volumes). Hence, you will need to run the update_assets target
+# everytime you update edx-platform sources and plan to develop in it.
+update-assets: tree create-symlinks  ## run update_assets to copy required statics in local volumes
```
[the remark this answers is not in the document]
sampaccoud: It's nice if we can keep short targets for make. It must be easy to type.
dev-assets, dev-watch ? With the make in front it makes a good semantic : make dev-assets, make dev-watch.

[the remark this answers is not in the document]
jmaupetit: With Makefile targets completion, having long target names is not a problem for me :wink: But, the naming schema you proposed seems great :clap:

config/lms/docker_run_production.py
```diff
@@ -675,7 +675,7 @@
 
 MONGODB_PASSWORD = config("MONGODB_PASSWORD", default="")
 MONGODB_HOST = config("MONGODB_HOST", default="mongodb")
-MONGODB_PORT = config("MONGODB_PORT", default=27017, formatter=int)
+MONGODB_PORT = int(config("MONGODB_PORT", default=27017))
```
jmaupetit: IMO the formatter should apply to environment variable **and** yaml settings, instead of hard casting this way.

config/lms/docker_run_production.py
```diff
@@ -642,7 +642,7 @@
 
 DATABASE_ENGINE = config("DATABASE_ENGINE", default="django.db.backends.mysql")
 DATABASE_HOST = config("DATABASE_HOST", default="mysql")
-DATABASE_PORT = config("DATABASE_PORT", default=3306, formatter=int)
+DATABASE_PORT = int(config("DATABASE_PORT", default=3306))
```
jmaupetit: As the mongo client is the only dependency requiring an integer for the server port setting, I wouldn't systematically cast every port setting to an integer.

PR #63, opened 2018-10-04
🔧(development) set Django session persistence to database

rmoch:
Default value is memory which is incompatible with frequent server restarts

### DIFF
--- a/config/cms/docker_run_development.py
+++ b/config/cms/docker_run_development.py
@@ -14,6 +14,7 @@ EMAIL_BACKEND = config(
     "EMAIL_BACKEND", default="django.core.mail.backends.console.EmailBackend"
 )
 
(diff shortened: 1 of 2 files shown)


sampaccoud:
It may be ok locally but if we run the development settings in OpenShift, we need to set the session backend to Redis.

jmaupetit:
> we need to set the session backend to Redis

In a docker-* context, shouldn't it be the default configuration?

jmaupetit:
Closing this as it seems outdated. I'm not sure why we would want to persist sessions in development but another way to achieve this globally (for development) would be to define this setting in the environment, _i.e._:

```ini
# env.d/development
SESSION_ENGINE=django.contrib.sessions.backends.cached_db
```

PR #74, opened 2018-11-20
➕(core) add a requirement file for Open edx Extended

sampaccoud:
## Purpose

This requirement file was originally located in the oee fork of edx-platform in openfun/edx-platform. 
In fact, the small fork in edx-platform works normally if the configurable LTI consumer is not installed.
It means the fork is robust and we can handle dependencies in openedx-docker. It is easier and makes more sense for all the dependencies that we need to add that have nothing to do with the edx-platform fork.

## Proposal

- [x] add a requirements.txt file to openedx-docker on the oee branch,
- [x] add basic requirements we know we need (configurable-lti-consumer and raven),
- [x] modify the Dockerfile to install requirements from this local file (we will then be able to remove the eponym file in edx-platform).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,10 @@
+# Open edX extended: dependencies
+
+# ==== xblocks ====
+
+configurable_lti_consumer-xblock==0.2.1
+
+
+# ==== third-party apps ====
+
+raven==6.9.0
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -0,0 +1,5 @@
+# Open edX extended: dependencies
+# ==== xblocks ====
+
+configurable_lti_consumer-xblock==0.2.1
+raven==6.9.0
```
jmaupetit: Raven is not an xblock :smile:

PR #79, opened 2018-12-11
✨(session) add redis as session backend

sampaccoud:
## Purpose

Redis is in our opinion the best practice to record sessions as it combines the speed of memory with a persistence layer.

## Proposal

- [x] add dependency to the `redis-session-backend` Django third-party app,
- [x] declare Redis as default session backend in lms and cms settings.


rmoch:
LGTM 👍

PR #80, opened 2018-12-17
🔧(cache) configure Django cache

rmoch:
Default configuration was using LocMemCache backend preventing
cache to be shared between containers.
By example, lms course plan was not invalidated when modifications
were made in studio.
We now use Memcached as default cache and LocMemcach for 'location mapping'
as recommended by edX
https://github.com/edx/edx-platform/blob/master/lms/envs/aws.py#L199

### DIFF
--- a/config/cms/docker_run_production.py
+++ b/config/cms/docker_run_production.py
@@ -148,9 +148,13 @@ LOG_DIR = config("LOG_DIR", default="/edx/var/logs/edx")
 CACHES = config(
     "CACHES",
     default={
-        "default": {
+        "loc_cache": {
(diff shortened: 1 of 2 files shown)


config/cms/docker_run_production.py
```diff
@@ -148,9 +148,13 @@
 CACHES = config(
     "CACHES",
     default={
-        "default": {
+        "loc_cache": {
```
sampaccoud: you don't need this cache in the cms.

rmoch: it is set in cms/aws.py

https://github.com/edx/edx-platform/blob/12ff3e116d9f98fd0504416918b671225c5970c2/cms/envs/aws.py#L168

PR #91, opened 2019-01-29
🐛(logging) fix wrong settings for tracking logs

rmoch:
We were initializing TRACKING_IGNORE_URL_PATTERNS to None instead of a list
of url to ignore when writing tracking logs, making each requests track a
TypeError.

### DIFF
--- a/config/cms/docker_run_production.py
+++ b/config/cms/docker_run_production.py
@@ -352,7 +352,7 @@ STUDIO_SHORT_NAME = config("STUDIO_SHORT_NAME", default=STUDIO_SHORT_NAME)
 
 # Event Tracking
 TRACKING_IGNORE_URL_PATTERNS = config(
-    "TRACKING_IGNORE_URL_PATTERNS", default=None, formatter=json.loads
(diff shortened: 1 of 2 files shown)


config/cms/docker_run_production.py
```diff
@@ -352,7 +352,7 @@
 
 # Event Tracking
 TRACKING_IGNORE_URL_PATTERNS = config(
-    "TRACKING_IGNORE_URL_PATTERNS", default=None, formatter=json.loads
+    "TRACKING_IGNORE_URL_PATTERNS", default=[r"^/event", r"^/login", r"^/heartbeat"], formatter=json.loads
```
sampaccoud: why do we want to ignore login?

jmaupetit: After thought: why not using `common.py` default value instead for maintainability?

```suggestion
    "TRACKING_IGNORE_URL_PATTERNS", default=TRACKING_IGNORE_URL_PATTERNS, formatter=json.loads
```

PR #92, opened 2019-01-31
🐛(logging) fix wrong settings for tracking logs

rmoch:
We were initializing TRACKING_IGNORE_URL_PATTERNS to None instead of a list
of url to ignore when writing tracking logs, making each requests track a
TypeError.


jmaupetit:
Thx @rmoch! :heart:

PR #104, opened 2019-03-19
WIP ✨(proctoring) add proctor-exam xblock to oee requirements

rmoch:
## Purpose

Deploy alpha version of newly created Proctor Exam xblock on D2RP instance for tests

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@
 # ==== xblocks ====
 
 configurable_lti_consumer-xblock==1.2.3
-
+-e git+https://github.com/openfun/xblock-proctor-exam.git@0.8.0-alpha#egg=xblock-proctor-exam
 
 # ==== third-party apps ====
 


requirements.txt
```diff
@@ -3,7 +3,7 @@
 # ==== xblocks ====
 
 configurable_lti_consumer-xblock==1.2.3
-
+-e git+https://github.com/openfun/xblock-proctor-exam.git@0.8.0-alpha#egg=xblock-proctor-exam
```
jmaupetit: This package should be published to PyPI and installed from it. We should not install dependencies this way when it can be avoided.

Moreover, I don't see any code or tags in the `master` branch of the project, this release comes out of the blue...

sampaccoud: @rmoch you can copy/paste what has been done on the [configurable LTI xblock](https://github.com/openfun/xblock-configurable-lti-consumer) or is it a very different case? The sandbox, CI and package is a best practice IMO so let's do the same WDYT?

PR #109, opened 2019-05-17
✨(edxapp) add Fonzie to OEE dependencies

rmoch:
## Purpose

Add Fonzie to add access control to instructor dashboard CSV export files, and to future edx-platform evolutions

### DIFF
--- a/config/lms/docker_run_production.py
+++ b/config/lms/docker_run_production.py
@@ -49,6 +49,9 @@ SESSION_REDIS = config(
     formatter=json.loads,
 )
 
+# Override edX LMS urls with Fonzie's
+ROOT_URLCONF = "lms.root_urls"
+
 # IMPORTANT: With this enabled, the server must always be behind a proxy that
 # strips the header HTTP_X_FORWARDED_PROTO from client requests. Otherwise,
 # a user can fool our server into thinking it was an https connection.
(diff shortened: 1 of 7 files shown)


config/lms/root_urls.py
```diff
@@ -0,0 +1,16 @@
...
+from __future__ import absolute_import, unicode_literals
+
+from django.conf.urls import include, url
+
+from lms.urls import urlpatterns  # pylint: disable=import-error
+
+
+urlpatterns += [
+    # Fonzie API urls
+    url(r"^api/", include("fonzie.urls", namespace="fonzie")),
```
jmaupetit: There is no edx urls prefixed with `api`?

rmoch: Indeed edX has many `^api/whatever` routes, however our `^api/v1.0/{status,acl}` will not interfer.
We may rename this `^fonzie/api/*` 
WDYT ?

jmaupetit: If our URK schema does not interfere, you can let it as is.

config/lms/docker_run_production.py
```diff
@@ -829,8 +832,20 @@
     "GRADES_DOWNLOAD_ROUTING_KEY", default=HIGH_MEM_QUEUE
 )
 
+# Fonzie API endpoint to add access control on instructor dashboard
+# CSV export files (activated by features ENABLE_GRADE_DOWNLOADS and
+# ALLOW_COURSE_STAFF_GRADE_DOWNLOADS)
+
```
jmaupetit: Empty line to remove

PR #114, opened 2019-09-09
🔥(images) remove timezone configuration step

jmaupetit:
## Purpose

One should not enforce a timezone in a Docker image. Instead, this should be set at the runtime depending on the host configuration or via the `TZ` environment variable for non-alpine images.

More, than that the configuration commands were buggy (or partially useless).

## Proposal

- [x] Remove timezone configuration from our `Dockerfile`s

### DIFF
--- a/releases/hawthorn/1/bare/Dockerfile
+++ b/releases/hawthorn/1/bare/Dockerfile
@@ -12,9 +12,12 @@ ARG EDX_RELEASE_REF=open-release/hawthorn.1
 # === BASE ===
 FROM ubuntu:16.04 as base
 
-# Configure locales
+# Configure locales & timezone
 RUN apt-get update && \
-    apt-get install -y gettext locales && \
+    apt-get install -y \
+      gettext \
+      locales \
+      tzdata && \
     rm -rf /var/lib/apt/lists/*
 RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
     locale-gen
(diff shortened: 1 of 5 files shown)


releases/hawthorn/1/bare/Dockerfile
```diff
@@ -185,14 +185,6 @@ COPY --from=builder /usr/local /usr/local
 # Copy modified sources (sic!)
 COPY --from=builder /edx/app/edxapp/edx-platform  /edx/app/edxapp/edx-platform
 
-# Set container timezone and related timezones database and DST rules
-# See https://serverfault.com/a/856593
-ENV TZ 'Etc/UTC'
```
lunika: `Etc/UTC` ? First time I see this timezone

jmaupetit:
> I don't know if the timezone is really configurable at runtime, even more if you are not a root user.

```
docker run --rm -e TZ="Etc/UTC" my-image
```

lunika:
> > I don't know if the timezone is really configurable at runtime, even more if you are not a root user.
> 
> ```
> docker run --rm -e TZ="Etc/UTC" my-image
> ```

It's enough ? There is no configuration to rebuild ? Are you sure Kevin ?

jmaupetit:
It is supported on most modern Linux distributions. Only alpine-based images need a reconfiguration.

jmaupetit:
You can test these commands as an example:

```
$ docker run --rm -e TZ=America/New_York debian:stretch date
Mon Sep  9 15:41:40 EDT 2019

$ docker run --rm -e TZ=UTC debian:stretch date
Mon Sep  9 19:41:49 UTC 2019
```

lunika:
> It is supported on most modern Linux distributions. Only alpine-based images need a reconfiguration.

I'm not modern at all, I still program using OOP... Thank you

PR #123, opened 2019-10-01
⬆️ (hawthorn/1/oee) upgrade xblock-poll to it's latest version

rmoch:
`xblock-poll` version 1.5.1 still used by latest versions of `edx-platform`
has a bug preventing instructor to download CSV files of user's
responses. Version 1.8.7 should fix that issue

### DIFF
--- a/releases/hawthorn/1/oee/requirements.txt
+++ b/releases/hawthorn/1/oee/requirements.txt
@@ -9,3 +9,6 @@ configurable_lti_consumer-xblock==1.2.3
 # ==== third-party apps ====
 raven==6.9.0
 django-redis-sessions==0.6.1
+
+# upgrade xblock-poll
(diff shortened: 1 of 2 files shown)


releases/hawthorn/1/oee/requirements.txt
```diff
@@ -9,3 +9,6 @@ configurable_lti_consumer-xblock==1.2.3
 # ==== third-party apps ====
 raven==6.9.0
 django-redis-sessions==0.6.1
+
+# upgrade xblock-poll
+git+https://github.com/open-craft/xblock-poll.git@v1.8.7#egg=django-openid-auth==1.8.7
```
jmaupetit: The egg name does not reflect the package name.

rmoch:
> If this bug affects all versions of edx-platform we build our images upon, you should also fix other releases/flavors.

right. yet still not sure this solves the issue

jmaupetit:
So you finally gave up?

PR #125, opened 2019-10-08
🔧(config) add LOCALE_PATHS to configurable settings

rmoch:
## Purpose

This sets and order paths when edx finds i18n files. Overriding
this settings allow us to override edx-platform i18n strings by
redefining them in theme.

### DIFF
--- a/releases/dogwood/3/bare/CHANGELOG.md
+++ b/releases/dogwood/3/bare/CHANGELOG.md
@@ -9,6 +9,10 @@ release.
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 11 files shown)


releases/master/bare/CHANGELOG.md
```diff
@@ -13,4 +13,8 @@ release.
 
 ## [Unreleased]
 
+### Changed
+
+- Add LOCALE_PATHS to configurable settings
+
```
jmaupetit: You don't need to add a change log entry for this flavor as we never release it (see the remark above, l.10-12).

PR #140, opened 2019-11-29
⚡️(circle) speed up new release publication

jmaupetit:
## Purpose 

When a new release has been tagged, we do not need to rebuild all releases and flavors we support. 

## Proposal

Only the hub job will get triggered upon tagging and it will only rebuild and publish the target release.


lunika:
Wow you forget the CHANGELOG !!!!

jmaupetit:
This was intended!

PR #143, opened 2019-12-03
Fix CI job names and execution

jmaupetit:
## Purpose

This project's CI needs a little bit more attention to be more efficient.

## Proposal

- [x] Avoid using slashes in job names so that CircleCI reports jobs properly in GitHub's pull requests
- [x] Avoid running all releases build/bootstrap when a PR only targets a single/few release(s)

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -eo pipefail
+
 # usage: display usage with the appropriate exit code
 #
 # usage: usage [EXIT_CODE]
@@ -14,6 +16,8 @@ function usage(){
 Available commands:
 
   activate_path    display CIRCLE_TAG (or CIRCLE_JOB) release activate script path
+  checkpoint       skip release job if changes are not targeting the current release
+  get_changes      get a list of changed files in the current branch compared to master
 "
 
     # shellcheck disable=SC2086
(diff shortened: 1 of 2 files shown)


bin/ci
```diff
@@ -13,41 +15,71 @@ function usage(){
...
   echo "releases/${release}/activate"
 }
 
+# Skip release job if changes are not targeting the current release
+function checkpoint(){
+
+  changes=$(get_changes)
+
+  if echo "${changes}" | grep -v "releases/" > /dev/null ; then
+    (>&2 echo "Current work scope is global, all releases should be build.")
```
sampaccoud:
```suggestion
    (>&2 echo "Current work scope is global, all releases should build.")
```

lunika:
You should maybe add an explanation in the readme on how to name our commits.
What will happen if I made several commits and each commit is targeting a specific flavor ? Only the last one will be tested ?

bin/ci
```diff
@@ -13,41 +15,71 @@ function usage(){
...
+    exit 0
+  fi
+
+  release_path=$(activate_path | sed -E 's|/activate$||')
```
sampaccoud: Just occured to me that `release` might be a confusing name after all. If I understand well, In this case, we are talking about a flavor, not a release (could be a job or could be a tag...).

So should we rename the `releases` folder to `flavors`?

jmaupetit: A "flavor" is a part of fully qualified release name. A flavor is the last leaf of a full qualification (that is translated to a FS path). IMO, in other words: a flavor is like an adjective, not a common name.

jmaupetit:
> You should maybe add an explanation in the readme on how to name our commits.

I don't see why it is related to this PR?

> What will happen if I made several commits and each commit is targeting a specific flavor ? Only the last one will be tested ?

Nope, we are detecting changes from a branch (PR) compared to `master`, so all releases with changes will be tested.

lunika:
> I don't see why it is related to this PR?

You can avoid my comment, I thought `$CIRCLE_JOB` was the commit title...

jmaupetit:
So. Are we good now?

bin/ci
sampaccoud: ok. But do you see the problem I raised about confusing a release (==tag) and a "distrib" or whatever we should call it?

jmaupetit: Since we discuss about it every time I submit a PR to this project, yes: I can see that there might be a confusion :smile:

PR #156, opened 2019-12-15
Dogwood: fix SSL certificate checking errors

jmaupetit:
## Purpose

Python2 openssl support in Ubuntu 12.04 relies on the outdated system openssl library, leading to SSL certicate verification failures (lack of SNI support, etc.). To mitigate this issue, it is recommended to install a recent release of urllib3 that uses pyOpenSSL instead to achieve the checking step.

## Proposal

- [x] upgrade `requests` to a recent release
- [x] install `urllib3[secure]`

### DIFF
--- a/releases/dogwood/3/bare/Dockerfile
+++ b/releases/dogwood/3/bare/Dockerfile
@@ -156,10 +156,19 @@ RUN pip install -r requirements/edx/base.txt
 RUN pip install -r requirements/edx/paver.txt
 RUN pip install -r requirements/edx/post.txt
 RUN pip install -r requirements/edx/local.txt
+# Install extras dependencies
+#
 # Redis is an extra requirement of Celery, we need to install it explicitly so
-# that celery workers are effective
-RUN pip install redis==3.3.7
-
+# that celery workers are effective.
+#
+# Requests should be upgraded with a recent secure flavor of urrllib3 to
+# prevent SSL certificate validation failure (we should use pyOpenSSL instead
+# of the local openssl library). For reference, see:
+# https://urllib3.readthedocs.io/en/latest/user-guide.html#ssl-py2
+RUN pip install --upgrade \
+      redis==3.3.7 \
+      requests==2.22.0 \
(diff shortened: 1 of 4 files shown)


releases/dogwood/3/bare/Dockerfile
```diff
@@ -156,10 +156,19 @@ RUN pip install -r requirements/edx/base.txt
...
-RUN pip install redis==3.3.7
-
+# that celery workers are effective.
+#
+# Requests should be upgraded with a recent secure flavor of urrllib3 to
+# prevent SSL certificate validation failure (we should use pyOpenSSL instead
+# of the local openssl library). For reference, see:
+# https://urllib3.readthedocs.io/en/latest/user-guide.html#ssl-py2
+RUN pip install --upgrade \
+      redis==3.3.7 \
```
sampaccoud: can't you remove this like you did in the Eucalyptus image?

releases/dogwood/3/bare/Dockerfile
```diff
@@ -156,10 +156,19 @@ RUN pip install -r requirements/edx/base.txt
...
+      requests==2.22.0 \
```
sampaccoud: I'm concerned by the possible side effects of this elsewhere. WDYT?

jmaupetit: I am also, but we have to test it and eventually downgrade it to a release that does not vendor `urllib3`. My bet is that if `requests` is following a semantic versioning convention, nothing should be broken in the API.

releases/dogwood/3/bare/Dockerfile
jmaupetit: Nope, I've removed it from `eucalyptus/3/wb` because it was already declared in the `requirements.txt` file, but we don't have such file (or step) in bare flavors.

releases/dogwood/3/bare/Dockerfile
sampaccoud: ok that's reassuring.

PR #159, opened 2019-12-15
🔧(*/flavors) configure `general` memcached cache backend

rmoch:
Edx code attempts to creates Memcached keys containing spaces which
is not allowed, so cache keys have to be sanitized.
Therefore default Django configuration is not suited, here we
set a `general` cache backend which is used by incorrect edx code and
provide a normalizing function.

### DIFF
--- a/releases/dogwood/3/bare/CHANGELOG.md
+++ b/releases/dogwood/3/bare/CHANGELOG.md
@@ -9,6 +9,11 @@ release.
 
 ## [Unreleased]
 
+### Added
+
+- Configure `general` cache backend including cache keys sanitizing function
+
+
 ## [dogwood.3-1.1.3] - 2019-12-15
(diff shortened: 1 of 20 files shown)


releases/dogwood/3/bare/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ release.
 
 ## [Unreleased]
 
+### Added
+
+- Configure `general` cache backend including sanitizing cache keys function
```
jmaupetit: I would suggest the following:
```suggestion
- Configure `general` cache backend including cache keys sanitizing function
```

rmoch: I rewrited commit message for clarity. Hope this is better

PR #166, opened 2019-12-20
✨([dogwood|eucalyptus]/3/[fun|wb]) activate studio's `videoupload` page

rmoch:
Both FUN flavors were missing some settings to correctly show up
`videoupload` page.

This is what it looks like if you wonder :-D
Capture d’écran 2019-12-20 à 15 16 56

### DIFF
--- a/releases/dogwood/3/fun/CHANGELOG.md
+++ b/releases/dogwood/3/fun/CHANGELOG.md
@@ -9,6 +9,10 @@ release.
 
 ## [Unreleased]
 
+### Fixed
+
+- Add missing `fun` application templates directory to default templates directories
+  to activate `videoupload` templates
 - Configure all cache backends as they are in FUN's production instance
 
 ## [dogwood.3-fun-1.6.0] - 2020-01-03
--- a/releases/eucalyptus/3/wb/CHANGELOG.md
+++ b/releases/eucalyptus/3/wb/CHANGELOG.md
@@ -9,9 +9,11 @@ release.
 
 ## [Unreleased]
 
-### Added
+### Fixed
 
 - Configure all cache backends as they are in FUN's production instance
+- Add missing `fun` application templates directory to default templates directories
+  to activate `videoupload` templates
 
 ### Changed
 
--- a/releases/dogwood/3/fun/config/cms/docker_run_production.py
+++ b/releases/dogwood/3/fun/config/cms/docker_run_production.py
@@ -783,6 +783,8 @@ GLOWBL_COLL_OPT = config("GLOWBL_COLL_OPT", default="FunMoocJdR")
 
 # This is dist-packages path where all fun-apps are
 FUN_BASE_ROOT = path(os.path.dirname(pkgutil.get_loader("funsite").filename))
+# Add to Mako template dirs path to `videoupload` panel templates
+DEFAULT_TEMPLATE_ENGINE["DIRS"].append(FUN_BASE_ROOT / "fun/templates/cms")
 
 # Add 'theme/cms/templates' directory to MAKO template finder to override some
 # CMS templates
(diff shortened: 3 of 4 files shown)


releases/eucalyptus/3/wb/config/cms/fun.py
```diff
@@ -45,9 +45,17 @@
 # CMS templates
 MAKO_TEMPLATES["main"] = [FUN_BASE_ROOT / "fun/templates/cms"] + MAKO_TEMPLATES["main"]
 
+# Add to Mako templates dirs path to `videoupload` panel templates
+DEFAULT_TEMPLATE_ENGINE["DIRS"].append(FUN_BASE_ROOT / "fun/templates/cms")
+
+# Display `videoupload` page in studio menus
+# This constant has to be not None in this flavor
+FUN_DEFAULT_VIDEO_PLAYER = config("FUN_DEFAULT_VIDEO_PLAYER", default=False, formatter=bool)
```
sampaccoud: @rmoch shouldn't you activate it by default?

By the way I checked in dogwood and it is not a boolean:
FUN_DEFAULT_VIDEO_PLAYER = "libcast_xblock"

rmoch: in dowood and eucalytpus settings, the use of this constant was removed in this recent commit: https://github.com/openfun/openedx-docker/commit/2300f8e746f5863f277beecef1d4fe54dac25ebc
see old code: https://github.com/openfun/fun-apps/blob/master/fun/envs/common.py#L156

It's now only used as a boolean by eucalyptus/wb template to display `videoupload` page: 
https://github.com/openfun/edx-platform/blob/fun/whitebrand/cms/templates/widgets/header.html#L70

Hope this help.

jmaupetit: So the default should be `True` instead of `False`? After looking at the code where it is used, it appears that `True` or `False` would lead to the same result as you are testing if it's not `None`...

releases/dogwood/3/fun/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ release.
 
 ## [Unreleased]
 
+### Fixed
+
+- Add missing template dir to `videoupload` templates
```
jmaupetit:
```suggestion
- Add missing `fun` application templates directory to default templates directories to activate `videoupload` templates
```

releases/eucalyptus/3/wb/CHANGELOG.md
```diff
@@ -37,6 +42,10 @@ release.
 ### Fixed
 
 - `AUTH_TOKENS` was wrongly used as a dictionary and default values were lost
+### Fixed
+
+- Add missing template dir to `videoupload` templates
```
jmaupetit: looks like a rebase failure

releases/dogwood/3/fun/config/cms/fun.py
```diff
@@ -0,0 +1,125 @@
+from glob import glob
```
sampaccoud: This file was removed and the FUN settings were added to docker_run_production.py see this commit in master: https://github.com/openfun/openedx-docker/commit/5b455a66e0bf54b589a61c9b88c6ace1f663ab63

rmoch: 👀

releases/eucalyptus/3/wb/CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ release.
 
 ## [Unreleased]
 
+### Added
```
sampaccoud: Should be Fixed like above?

PR #172, opened 2020-01-06
♻️(flavors/*) make ORA2 configurable on all flavors

sampaccoud:
## Purpose

We want to make ORA2 work on all flavors. Except for `dogwood.3-fun`, the default backend should be set to the filesystem but everything should be configurable via yaml configuration files.

## Proposal

- [x] add a cache backend dedicated to ORA2 using memcached like the default cache but on which we set a specific key prefix,
- [x] add missing settings to each flavor with a `config` getter,
- [x] clean-up the configuration of the `dogwood.3-fun` image which uses Swift.

### DIFF
--- a/releases/dogwood/3/bare/CHANGELOG.md
+++ b/releases/dogwood/3/bare/CHANGELOG.md
@@ -15,6 +15,7 @@ release.
 
 ### Changed
 
+- Make ORA2 configurable and use filesystem backend by default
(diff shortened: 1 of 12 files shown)


jmaupetit:
@sampaccoud don't forget the changelog! 

...and releases?

sampaccoud:
@rmoch I'm not changing anything to dogwood.3-fun. Just cosmetics and removing a useless setting.

PR #175, opened 2020-01-08
Release dogwood.3 fun 1.6.1 and eucalyptus.3 wb 1.5.0

sampaccoud:
### Fixed

- Add missing `fun` application templates directory to default templates directories
  to activate `videoupload` templates
- Configure all cache backends as they are in FUN's production instance

### Fixed

- Configure all cache backends as they are in FUN's production instance
- Add missing `fun` application templates directory to default templates directories
  to activate `videoupload` templates

### Changed

- Make ORA2 configurable and use filesystem backend by default

### DIFF
--- a/releases/dogwood/3/fun/CHANGELOG.md
+++ b/releases/dogwood/3/fun/CHANGELOG.md
@@ -9,6 +9,8 @@ release.
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


sampaccoud:
Sorry I went too fast. I need to merge PR https://github.com/openfun/openedx-docker/pull/174 first.

PR #177, opened 2020-01-09
🐛(flavors/*) Ensure all required directories exists inside each volume

sampaccoud:
## Purpose

Some required directories were missing inside volumes. It seems that edX does not ensure this by code when using these locations.

### Proposal

Add a check in setetings to make sure all required directories exist upon startup.

This was ported from [fun-apps](https://github.com/openfun/fun-apps/blob/0af5fab8ae6967428bb0a44568bdd84bccbbfcea/fun/envs/common.py#L238)


sampaccoud:
ok

PR #181, opened 2020-01-11
🐛([dogwood|eucalyptus]/3/[fun|wb]) fix videofront subtitles cache

sampaccoud:
## Purpose

The location declared for videofront subtitles cache did not exist which raised 500 errors when trying to access it.

## Proposal

- [x] add our now beloved anti-pattern to ensure the required directory is created in the volume
- [x] move videofront cache declaration to the "FUN stuff" section of the settings so that we don't forget to remove it when we soon decommission videofront.


sampaccoud:
@jmaupetit the CHANGELOG already contained an item that corresponds to what we are doing here: 
> Ensure all required directories exists inside each volume

PR #184, opened 2020-01-15
Add edxapp-nginx extras

jmaupetit:
## Purpose

We need a way to bundle an openedx release assets with the `nginx` image that will serve them to avoid side-effects in our deployment procedure (see the Arnold project).

## Proposal

Current work copy files collected in the edxapp release and replace duplicates by symlinks to save disk space and network bandwidth (up to 65%):

```
dogwood.3-1.2.1          	529MB
dogwood.3-1.2.1-slim     	198MB
eucalyptus.3-1.1.1       	573MB
eucalyptus.3-1.1.1-slim  	222MB
hawthorn.1-3.0.0         	959MB
hawthorn.1-3.0.0-slim    	320MB
hawthorn.1-oee-3.0.0     	960MB
hawthorn.1-oee-3.0.0-slim	320MB
```
*-slim images have been generated with the aforementioned optimization.

During this work, I've also:

- [x] upgrade the CI `docker-compose` release to `1.25.1`
- [x] fix `eucalyptus/3/wb` git branch reference

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,14 @@ EDX_RELEASE_REF           ?= release-2018-08-29-14.14
 EDX_DEMO_RELEASE_REF      ?= master
 EDX_DEMO_ARCHIVE_URL      ?= https://github.com/edx/edx-demo-course/archive/$(EDX_DEMO_RELEASE_REF).tar.gz
 
+# Docker images
+EDXAPP_IMAGE_NAME         ?= edxapp
+EDXAPP_IMAGE_TAG          ?= $(EDX_RELEASE)-$(FLAVOR)
+
+# Extras
+EXTRAS_NGINX_IMAGE_NAME   ?= edxapp-nginx
+EXTRAS_NGINX_IMAGE_TAG    ?= $(EDXAPP_IMAGE_TAG)
+
 # Redis service used
 REDIS_SERVICE             ?= redis
 
@@ -21,6 +29,7 @@ COMPOSE          = \
   DOCKER_UID=$(DOCKER_UID) \
   DOCKER_GID=$(DOCKER_GID) \
   FLAVORED_EDX_RELEASE_PATH="$(FLAVORED_EDX_RELEASE_PATH)" \
+  EDXAPP_IMAGE_TAG=$(EDXAPP_IMAGE_TAG) \
   docker-compose
 COMPOSE_RUN      = $(COMPOSE) run --rm -e HOME="/tmp"
 COMPOSE_EXEC     = $(COMPOSE) exec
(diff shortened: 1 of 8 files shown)


sampaccoud:
The question is: do we really need the "non-slim" images... is slim not the one and only edxapp-nginx image!?

extras/nginx/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN apt-get update && \
+    apt-get install -y rdfind && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN python manage.py lms collectstatic --noinput --settings=fun.docker_run && \
+    python manage.py cms collectstatic --noinput --settings=fun.docker_run
+
+# Replace duplicated file by a symlink to decrease the overall size of the
+# final image
+RUN rdfind -makesymlinks true /edx/app/edxapp/staticfiles
```
sampaccoud: https://media.giphy.com/media/NmerZ36iBkmKk/giphy.gif

jmaupetit:
> The question is: do we really need the "non-slim" images... is slim not the one and only edxapp-nginx image!?

Of course we won't build or use non-slim image! This table's purpose is to quantify the image size variation with and without `rdfind` usage.

extras/nginx/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+# === NGINX (final image) ===
+FROM ${NGINX_IMAGE_NAME}:${NGINX_IMAGE_TAG}
+
+ARG EDXAPP_STATIC_ROOT
+
+RUN mkdir -p /edx/app/edxapp
```
sampaccoud: The path of the directory to create should depend on EDXAPP_STATIC_ROOT. 
Actually you don't need to variabilize the location where you put the files in the nginx container... only the location of the files to which they will be in the edxapp image?

jmaupetit: Because of the symlinks (absolute paths) the nginx image statics path should be identical to their original location in edxapp image.

sampaccoud: I see. Then you should still variabilize the location!
```suggestion
RUN mkdir -p "$EDXAPP_STATIC_ROOT"
```

Makefile
```diff
@@ -229,6 +233,16 @@ dev-watch: tree  ## start assets watcher (front-end development)
 	  paver watch_assets --settings=fun.docker_build_development
 .PHONY: dev-watch
 
+extras-build:  ## build all project extras images for a flavored release
+	@echo -e "🐳 Building extras: $(COLOR_INFO)$(EXTRAS_NGINX_IMAGE):$(EXTRAS_NGINX_TAG)$(COLOR_RESET) image..."
+	cd extras/nginx && \
+	  docker build \
+	    --build-arg EDXAPP_IMAGE_NAME=edxapp \
```
lunika: you should create an environment variable to define `EDXAPP_IMAGE_NAME` arg value

jmaupetit:
@lunika I've added a fixup commit for your comment. Feel free to review it once again :pray:

PR #187, opened 2020-01-20
🐛(dogwood/3/fun) Fix known unpacking bug in xmodule

sampaccoud:
### Purpose

A long standing and known bug was fixed in [openfun/edx-platform](https://github.com/openfun/edx-platform/releases/tag/dogwood.3-fun-5.3.1).

### Proposal

Upgrade to openfun/edx-platform version dogwood.3-fun-5.3.1

:warning: The release process was refactored in openfun/edx-platform so the new tags are more explicit.

### DIFF
--- a/releases/dogwood/3/fun/Dockerfile
+++ b/releases/dogwood/3/fun/Dockerfile
@@ -9,8 +9,8 @@
 #     .
 ARG DOCKER_UID=1000
 ARG DOCKER_GID=1000
-ARG EDX_RELEASE_REF=fun/dogwood
-ARG EDX_ARCHIVE_URL=https://github.com/openfun/edx-platform/archive/fun-4.35.0.tar.gz
+ARG EDX_RELEASE_REF=dogwood.3-fun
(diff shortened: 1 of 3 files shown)


releases/dogwood/3/fun/Dockerfile
```diff
@@ -10,7 +10,7 @@
 ARG DOCKER_UID=1000
 ARG DOCKER_GID=1000
 ARG EDX_RELEASE_REF=fun/dogwood
-ARG EDX_ARCHIVE_URL=https://github.com/openfun/edx-platform/archive/fun-4.35.0.tar.gz
+ARG EDX_ARCHIVE_URL=https://github.com/openfun/edx-platform/archive/dogwood.3-fun-5.3.1.tar.gz
```
jmaupetit: I've fixed `eucalyptus.3-wb` in #184. You should also fix the `activate` shell script. For reference, see: ff435b1

sampaccoud: @jmaupetit Well spotted. It should be ok now.

PR #193, opened 2020-01-23
Upgrade fun-apps in wb + fun flavors

jmaupetit:
## Purpose

We recently fixed a cache busting issue on fun-apps' static files:

- https://github.com/openfun/fun-apps/pull/677
- https://github.com/openfun/fun-apps/pull/678

## Proposal

- [x] upgrade fun-apps to `2.2.1+wb` for the `eucalyptus.3-wb` release
- [x] upgrade fun-apps to `5.2.2` for the `dogwood.3-fun` release

### DIFF
--- a/releases/dogwood/3/fun/CHANGELOG.md
+++ b/releases/dogwood/3/fun/CHANGELOG.md
@@ -9,6 +9,8 @@ release.
 
 ## [Unreleased]
 
+## [dogwood.3-fun-1.8.6] - 2020-01-23
+
 ### Fixed
 
 - Upgrade `fun-apps` to `5.2.2`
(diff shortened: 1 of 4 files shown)


releases/dogwood/3/fun/CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ release.
 
 ## [Unreleased]
 
+## [dogwood.3-fun-1.8.6] - 2020-01-23
+
+### Fixed
```
lunika: Why not using the `Changed` section ?

jmaupetit: I hesitated, my conclusion was that it's a patch release and we are upgrading fun-apps to a patch release.

lunika: I am not convinced. But according to semver, we bump a patch if we have a fix... So maybe it's the good section

PR #194, opened 2020-01-23
Fix hawthorn.1 expected webpack-stats.json files

jmaupetit:
## Purpose

Now that we may use `hawthorn` images without mounting a dedicated volume for static files (see the edxapp-nginx extra), the backend still requires `webpack-stats.json` files generated via the `update_assets` `paver` command.

## Proposal

- [x] copy `webpack-stats.json` files where expected in the production image

### DIFF
--- a/releases/hawthorn/1/bare/Dockerfile
+++ b/releases/hawthorn/1/bare/Dockerfile
@@ -205,6 +205,14 @@ COPY --from=builder /usr/local /usr/local
 # Copy modified sources (sic!)
 COPY --from=builder /edx/app/edxapp/edx-platform  /edx/app/edxapp/edx-platform
 
+# ... and webpack-stats.json files (sic! ...again)
+COPY --from=builder \
+  /edx/app/edxapp/edx-platform/common/static/webpack-stats.json \
+  /edx/app/edxapp/staticfiles/webpack-stats.json
+COPY --from=builder \
+  /edx/app/edxapp/edx-platform/common/static/studio/webpack-stats.json \
+  /edx/app/edxapp/staticfiles/studio/webpack-stats.json
+
 # Now that dependencies are installed and configuration has been set, the above
 # statements will run with a un-privileged user.
 USER 10000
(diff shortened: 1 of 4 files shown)


releases/hawthorn/1/bare/Dockerfile
```diff
@@ -205,6 +205,14 @@ COPY --from=builder /usr/local /usr/local
 # Copy modified sources (sic!)
 COPY --from=builder /edx/app/edxapp/edx-platform  /edx/app/edxapp/edx-platform
 
+# ... and webpack-stats.json files (sic! ...again)
```
sampaccoud: you should then be able to remove what is at line 125: https://github.com/openfun/openedx-docker/blob/315f41da257aafcd0069db0558973a6db85dfd25/releases/hawthorn/1/bare/Dockerfile#L125

releases/hawthorn/1/oee/Dockerfile
```diff
@@ -212,6 +212,14 @@ COPY --from=builder /usr/local /usr/local
 # Copy modified sources (sic!)
 COPY --from=builder /edx/app/edxapp/edx-platform  /edx/app/edxapp/edx-platform
 
+# ... and webpack-stats.json files (sic! ...again)
```
sampaccoud: ditto

releases/hawthorn/1/bare/Dockerfile
jmaupetit: :thinking: I don't get your point: my fix does not fix this.

sampaccoud: ok so as discussed IRL we should leave it for people who will not use `edxapp-nginx` image for staics.

jmaupetit: As discussed IRL, we won't force the `edxapp-nginx` usage and thus this issue is not fixed.

jmaupetit: Concurrency my love... :smile:

PR #208, opened 2020-03-19
🔧([dogwood|eucalyptus]/3/[fun|wb]) add AT_INTERNET_SMARTTAG settings

rmoch:
This settings will be used in templates to configure
AT Internet web analytics tracking.


sampaccoud:
I would say it should only be set in your project. Probably in the docker_run.py file of your arnold configuration files.

jmaupetit:
I do agree with @sampaccoud: this is too customer-specific to stand in the default settings.

rmoch:
ok then 👍

PR #210, opened 2020-03-24
Add rate limit backend for proxy

sampaccoud:
## Purpose

Due to very heavy traffic during Coronavirus confinement , we found out that Open edX rate limiting authentication backend does not work properly behind a proxy. Indeed, it is using the `REMOTE_ADDR` header to determine a request's end user IP address. 

When placed behind a proxy, the actual end user IP address is found in the `X-Forwarded-For` header.

## Proposal

- Add a custom rate limiting authentication backend based on the X-Forwarded-For header,
- Refactor the way authentication backends are defined to make it more straightforward.

### DIFF
--- a/releases/dogwood/3/bare/CHANGELOG.md
+++ b/releases/dogwood/3/bare/CHANGELOG.md
@@ -13,6 +13,10 @@ release.
 
 - Rate limiting authentication backend that works behind proxies
 
+### Changed
+
+- Refactor the way authentication backends are configured to make it straightforward
+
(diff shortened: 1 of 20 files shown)


sampaccoud:
@jmaupetit Should I? I'm not sure since it is not available at all in Open edX... I don't know how it works for other people :man_shrugging:

jmaupetit:
@sampaccoud IMO it should be the default at least for the `dogwood.3-fun` flavor. WDYT?

sampaccoud:
@jmaupetit ok good idea.

PR #219, opened 2020-04-07
WIP 🐛(flavors/*) set basic authentification backend for dev. environments

rmoch:
In a previous modification we changed rate limit authentification
backend to work behind proxies and it brokes development images because
it can't work with Django development server as it does not set
HTTP_X_FORWARDED_FOR header.

### DIFF
--- a/releases/dogwood/3/bare/config/lms/docker_run_development.py
+++ b/releases/dogwood/3/bare/config/lms/docker_run_development.py
@@ -1,6 +1,8 @@
 # This file includes overrides to build the `development` environment for the LMS starting from the
 # settings of the `production` environment
 
+import json
+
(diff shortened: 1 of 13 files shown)


releases/dogwood/3/bare/config/lms/docker_run_development.py
```diff
@@ -21,3 +21,5 @@
 STATICFILES_STORAGE = "openedx.core.storage.DevelopmentStorage"
 
 ALLOWED_HOSTS = ["*"]
+
+AUTHENTICATION_BACKENDS = ['django.contrib.auth.backends.ModelBackend']
```
sampaccoud, line 25: I think you should set a default but still make it configurable.

PR #223, opened 2020-04-14
Refactor static files generation to avoid collectstatic during deployment

sampaccoud:
## Purpose

We recently [introduced a custom nginx-edxapp image](https://github.com/openfun/openedx-docker/pull/184)  that directly holds static files in order to serve them more efficiently with Kubernetes.

By doing this, we also thought we could speed up deployment because we would not need to run the `collectstatic` command upon each deployment. This last point was not successful because Open edX is using the `CachedStaticFilesStorage` by default, which needs to access static files in order to calculate their hash at runtime, when the application first encounters them.

Using the `ManifestStaticFilesStorage` backend should solve this issue and allow us to remove the static volume and stop running the `collectstatic` command upon deployment.

## Proposal

- [x] Add new static files storage backends for the lms and the cms, based on Django's `ManifestStaticFilesStorage` backend instead of `CachedStaticFilesStorage`,
- [x] Run the `collectstatic` command during image build and save only the manifest files to the production image,
- [x] In `docker-compose`, stop mounting the static directory in edxapp containers, in order to make sure that the application now works without accessing its static files,
- [x] Allow placing static files behind a CDN (while still using nginx as the origin).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,8 +155,8 @@ collectstatic: \
   tree \
   run
 collectstatic:  ## copy static assets to static root directory
-	$(MANAGE_LMS) collectstatic --noinput --settings=fun.docker_run
-	$(MANAGE_CMS) collectstatic --noinput --settings=fun.docker_run
+	$(COMPOSE_RUN) -v $(PWD)/$(FLAVORED_EDX_RELEASE_PATH)/data/static/production:/edx/app/edxapp/staticfiles lms python manage.py lms collectstatic --noinput --settings=fun.docker_run
+	$(COMPOSE_RUN) -v $(PWD)/$(FLAVORED_EDX_RELEASE_PATH)/data/static/production:/edx/app/edxapp/staticfiles cms python manage.py cms collectstatic --noinput --settings=fun.docker_run
 .PHONY: collectstatic
 
 create-symlinks:  ## create symlinks to local configuration (mounted via a volume)
--- a/releases/dogwood/3/fun/config/cms/storage.py
+++ b/releases/dogwood/3/fun/config/cms/storage.py
@@ -1,0 +1,28 @@
+"""
+Django file storage backends for the CMS that works behind a CDN.
+"""
+from django.conf import settings
+from django.contrib.staticfiles.storage import ManifestStaticFilesStorage
+
+from pipeline.storage import PipelineMixin
+
+
+class CMSManifestStaticFilesStorage(
+    PipelineMixin, ManifestStaticFilesStorage
+):
+    """
+    This static files storage backend works when static files are in the `edxapp-nginx` image,
+    eventually placed behing a CDN.
+
+    PipelineMixin: https://github.com/jazzband/django-pipeline
+    """
+
+    def url(self, name, force=False):
+        """Prepend static files path by the CDN base url when configured in settings."""
+        url = super(CMSManifestStaticFilesStorage, self).url(name, force=force)
+
+        cdn_base_url = getattr(settings, "CDN_BASE_URL", None)
+        if cdn_base_url:
+            url = "{:s}{:s}".format(cdn_base_url, url)
+
+        return url
--- a/releases/dogwood/3/fun/config/lms/storage.py
+++ b/releases/dogwood/3/fun/config/lms/storage.py
@@ -1,6 +1,7 @@
 """
 Django file storage backend for the LMS that works behind a CDN.
 """
+from django.conf import settings
 from django.contrib.staticfiles.storage import ManifestStaticFilesStorage
 
 from pipeline.storage import PipelineMixin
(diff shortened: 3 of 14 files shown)


releases/dogwood/3/fun/config/cms/storage.py
```diff
@@ -0,0 +1,28 @@
...
+    PipelineMixin: https://github.com/jazzband/django-pipeline
+    """
+
+    def url(self, name, force=False):
+        """Prepend static files path by the CDN base url when configured in settings."""
+        url = super(CMSManifestStaticFilesStorage, self).url(name, force=force)
+
+        cdn_base_url = getattr(settings, "CDN_BASE_URL", None)
+        if cdn_base_url:
+            url = "{:s}{:s}".format(cdn_base_url, url)
```
madmatah, line 26:
```suggestion
            url = f"{cdn_base_url}{url}"
```

releases/dogwood/3/fun/config/lms/storage.py
```diff
@@ -0,0 +1,31 @@
...
+    PipelineMixin: https://github.com/jazzband/django-pipeline
+    """
+
+    def url(self, name, force=False):
+        """Prepend static files path by the CDN base url when configured in settings."""
+        url = super(LMSManifestStaticFilesStorage, self).url(name, force=force)
+
+        cdn_base_url = getattr(settings, "CDN_BASE_URL", None)
+        if cdn_base_url:
+            url = "{:s}{:s}".format(cdn_base_url, url)
```
madmatah, line 28:
```suggestion
            url = f"{cdn_base_url}{url}"
```

releases/dogwood/3/fun/config/cms/storage.py
```diff
@@ -0,0 +1,28 @@
...
+
+from pipeline.storage import PipelineMixin
+
+
+class CMSManifestStaticFilesStorage(
+    PipelineMixin, ManifestStaticFilesStorage
+):
+    """
+    This static files storage backend works when static files are in the `edxapp-nginx` image,
+    eventually placed behing a CDN.
```
rmoch, line 15:
```suggestion
    possibly placed behing a CDN.
```

releases/dogwood/3/fun/config/cms/storage.py
jmaupetit: We are running python 2.7 ...

releases/dogwood/3/fun/config/lms/storage.py
jmaupetit: ditto

Makefile
```diff
@@ -155,8 +155,8 @@ collectstatic: \
   tree \
   run
 collectstatic:  ## copy static assets to static root directory
-	$(MANAGE_LMS) collectstatic --noinput --settings=fun.docker_run
-	$(MANAGE_CMS) collectstatic --noinput --settings=fun.docker_run
+	$(COMPOSE_RUN) -v $(PWD)/$(FLAVORED_EDX_RELEASE_PATH)/data/static/production:/edx/app/edxapp/staticfiles lms python manage.py lms collectstatic --noinput --settings=fun.docker_run
+	$(COMPOSE_RUN) -v $(PWD)/$(FLAVORED_EDX_RELEASE_PATH)/data/static/production:/edx/app/edxapp/staticfiles cms python manage.py cms collectstatic --noinput --settings=fun.docker_run
 .PHONY: collectstatic
```
jmaupetit, line 160: I might be wrong, but since i. static files have already been collected in the (production) image, ii. static files volume is no longer mounted for production services, and iii. static file volumes are not used in dev services (while using the development server), I think we can safely remove this no longer useful rule. WDYT?

jmaupetit: :thinking: after thought: this step is still required for all flavors except `dogwood.3-fun`...

releases/dogwood/3/fun/config/cms/storage.py
madmatah: oops.

Makefile
sampaccoud: No, the static files are collected in the build but only to save the manifest.
Static files still need to be collected to the volume mounted in the nginx image.
We could remove it if we used the nginx-edxapp image in docker-compose... but the way it is done with a standard nginx image is also a valid way to deploy openedx-docker...

You are right that merging this would break the other flavors... so if it is ok, I will do the same on all flavors before merging it...

releases/dogwood/3/fun/config/cms/storage.py
sampaccoud: ok. Why don't you like the word eventually?

rmoch: https://anglais-pratique.fr/index.php/rubriques/faux-amis/69-eventuellement-eventually
If I correctly understood the purpose of this PR, a CDN is an available option, not mandatory.

sampaccoud: ok !! faux ami. Well done.

sampaccoud:
This approach is abandoned in favor of PR https://github.com/openfun/openedx-docker/pull/229

PR #227, opened 2020-04-20
Hawthorn oee redis cache

rmoch:
Redis/Redis Sentinel configuration for Hawthorn OEE flavor

### DIFF
--- a/releases/hawthorn/1/oee/requirements.txt
+++ b/releases/hawthorn/1/oee/requirements.txt
@@ -8,6 +8,7 @@ configurable_lti_consumer-xblock==1.2.3
 
 # ==== third-party apps ====
 celery-redis-sentinel==0.3.0
+django-redis-sentinel-redux==0.2.0
 django-redis-sessions==0.6.1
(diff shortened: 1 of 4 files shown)


releases/hawthorn/1/oee/requirements.txt
```diff
@@ -8,6 +8,7 @@ configurable_lti_consumer-xblock==1.2.3
 
 # ==== third-party apps ====
 celery-redis-sentinel==0.3.0
+django-redis-sentinel-redux==0.2.0
```
sampaccoud, line 11: so you don't need this here?
```
django-redis==4.8.0  # Force django-redis 4.8.0 to keep Django==1.8
```

rmoch: No, Hawthorn is running `Django 1.11`. `django-redis-sentinel-redux` which requires `django-redis >= 4.5` will install `django-redis` 4.10 which is compatible with `Django 1.11` (but not with 1.8)

PR #229, opened 2020-05-13
Run collectstatic in the image with link option

sampaccoud:
## Purpose

We recently introduced a custom `edxapp-nginx` image that includes all static files and makes serving static files more efficient (direct from the Kubernetes node instead of via a PVC).

But `edxapp` still needed to have access to all static files in order to calculate the files' hash and compute their url. Since the static files were not collected inside the `edxapp` image, we had to mount a staticfiles PVC just for this, and run `collectstatic` upon each deployment... very slow!

## Proposal

This PR proposes to run `collectstatic` in the `edxapp` image so that all static files are available at run time.

We didn't want the image size to increase significantly, so the `collectstatic` command is run with the `--link` option, so that the files are not copied but symlinked. The manifest version of the files are still symlinked by running `rdfind`.

I also propose to run the custom `edxapp-nginx` image in the Docker Compose project since it is the preferred way to run `edxapp` in Docker, and one that we want to test in this project.


README.md
```diff
@@ -6,9 +6,6 @@ install a complete and customized version of [Open edX](https://open.edx.org).
 The idea is to handcraft a `Dockerfile`, in order to make the project simpler,
 more flexible and fully operable by developers.
 
-We hope this kind of `Docker` configuration can soon be included in the
-edx-platform repository itself.
```
jmaupetit, line 10: :sweat_smile:

PR #238, opened 2020-08-31
🐛(all flavors) fix redis-sentinel support

madmatah:
## Purpose

We want to use redis-sentinel as a cache backend in openedx-docker (all flavors).

As documented in the README, we should be able to do it by setting `export REDIS_SERVICE=redis-sentinel`.

But it does not work, the bootstrap of the project fails because the cache settings does not take this into account.

On top of that, on `hawthorn` and `dogwood`, the installed  `django-redis` dependency is not compatible with `django-redis-sentinel-redux`.

## Proposal

- Dogwood
  - [x] Fix settings to take `REDIS_SERVICE` into account

- Hawthorn
  - [x] Fix settings to take `REDIS_SERVICE` into account
  - [x] Pin `django-redis` to 4.5.0

- Eucalyptus
  - [x] Fix settings to take `REDIS_SERVICE` into account
  - [x] Pin `django-redis` to 4.5.0

### DIFF
--- a/releases/dogwood/3/fun/config/cms/docker_run_production.py
+++ b/releases/dogwood/3/fun/config/cms/docker_run_production.py
@@ -167,53 +167,41 @@ LOG_DIR = config("LOG_DIR", default=path("/edx/var/logs/edx"), formatter=path)
 CACHE_REDIS_HOST = config("CACHE_REDIS_HOST", default="redis")
 CACHE_REDIS_PORT = config("CACHE_REDIS_PORT", default=6379, formatter=int)
 CACHE_REDIS_DB = config("CACHE_REDIS_DB", default=1, formatter=int)
-CACHE_REDIS_URI = "redis://{}:{}/{}".format(CACHE_REDIS_HOST, CACHE_REDIS_PORT, CACHE_REDIS_DB)
-CACHE_REDIS_BACKEND = config("CACHE_REDIS_BACKEND", default="django_redis.cache.RedisCache")
-CACHE_REDIS_CLIENT = config("CACHE_REDIS_CLIENT", default="django_redis.client.DefaultClient")
+CACHE_REDIS_BACKEND = config(
+    "CACHE_REDIS_BACKEND", default="django_redis.cache.RedisCache"
+)
+CACHE_REDIS_URI = "redis://{}:{}/{}".format(
+    CACHE_REDIS_HOST, CACHE_REDIS_PORT, CACHE_REDIS_DB
+)
+CACHE_REDIS_CLIENT = config(
+    "CACHE_REDIS_CLIENT", default="django_redis.client.DefaultClient"
+)
+
+CACHES_DEFAULT_CONFIG = {
+    "BACKEND": CACHE_REDIS_BACKEND,
+    "LOCATION": CACHE_REDIS_URI,
+    "OPTIONS": {"CLIENT_CLASS": CACHE_REDIS_CLIENT},
+}
+
+if "Sentinel" in CACHE_REDIS_BACKEND:
+    CACHES_DEFAULT_CONFIG["LOCATION"] = [(CACHE_REDIS_HOST, CACHE_REDIS_PORT)]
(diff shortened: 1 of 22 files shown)


lunika (approved):

jmaupetit (approved):

releases/dogwood/3/fun/config/cms/docker_run_production.py
```diff
@@ -167,53 +167,36 @@
 CACHE_REDIS_HOST = config("CACHE_REDIS_HOST", default="redis")
 CACHE_REDIS_PORT = config("CACHE_REDIS_PORT", default=6379, formatter=int)
 CACHE_REDIS_DB = config("CACHE_REDIS_DB", default=1, formatter=int)
-CACHE_REDIS_URI = "redis://{}:{}/{}".format(CACHE_REDIS_HOST, CACHE_REDIS_PORT, CACHE_REDIS_DB)
 CACHE_REDIS_BACKEND = config("CACHE_REDIS_BACKEND", default="django_redis.cache.RedisCache")
+CACHE_REDIS_URI = "redis://{}:{}/{}".format(CACHE_REDIS_HOST, CACHE_REDIS_PORT, CACHE_REDIS_DB)
 CACHE_REDIS_CLIENT = config("CACHE_REDIS_CLIENT", default="django_redis.client.DefaultClient")
 
+CACHES_DEFAULT_CONFIG = {"BACKEND": CACHE_REDIS_BACKEND, "LOCATION": CACHE_REDIS_URI,
```
slandrault, line 174: [PEP8] Wouldn't it be cleaner to write that on multiple lines like so:

```suggestion
CACHES_DEFAULT_CONFIG = {
    "BACKEND": CACHE_REDIS_BACKEND,
    "LOCATION": CACHE_REDIS_URI,
    "OPTIONS": { "CLIENT_CLASS": CACHE_REDIS_CLIENT }
}
```

madmatah: That's right, thanks ! I forgot to run [black](https://github.com/psf/black) on the files.
I'll fix that.

madmatah: Done :heavy_check_mark:

releases/eucalyptus/3/wb/config/lms/docker_build_production.py
```diff
@@ -5,3 +5,35 @@
 DATABASES = {"default": {}}
 
 XQUEUE_INTERFACE = {"url": None, "django_auth": None}
+
+CACHES = {
```
sampaccoud, line 9: I'm wondering whether settings CACHES to just a default cache in development would be enough to make Django fall back to the default cache for all caches? It should work if the cache backend is userd properly in the code base.

releases/eucalyptus/3/wb/requirements.txt
```diff
@@ -12,7 +12,7 @@ xblock-utils2==0.3.0
 
 # ==== third-party apps ====
 celery-redis-sentinel==0.3.0
-django-redis==4.8.0  # Force django-redis 4.8.0 to keep Django==1.8
+django-redis==4.5.0
```
sampaccoud, line 15: Add comment about the necessary pinning?

releases/eucalyptus/3/wb/config/lms/docker_build_production.py
madmatah: I also think it should work. But since I don't know how cache keys are generated by the application (and therefore if there is a risk of collision), I prefered not to introduce some hazardous changes in this PR.

releases/eucalyptus/3/wb/requirements.txt
madmatah: Done, thanks!

releases/eucalyptus/3/wb/config/lms/docker_build_production.py
sampaccoud: fair enough

sampaccoud (approved):

PR #240, opened 2020-09-01
🎨(bin) sort release in reverse order for bin/activate

slandrault:
## Purpose

In order to easily identify the history of the different release
when running the bin/activate command, a sort operation is added
to have the most recent branches first.

## Proposal

Adding a `sort` command enables to sort the list of release (in reverse order) to have the most recent release first (sorted by alphabetical order)

Before (unordered):
```plain
Select an available flavored release to activate:
[1] master/bare (default)
[2] dogwood/3/fun
[3] dogwood/3/bare
[4] ironwood/2/bare
[5] eucalyptus/3/wb
[6] eucalyptus/3/bare
[7] hawthorn/1/oee
[8] hawthorn/1/bare
```

After (most recent first):
```plain
Select an available flavored release to activate:
[1] master/bare (default)
[2] ironwood/2/bare
[3] hawthorn/1/oee
[4] hawthorn/1/bare
[5] eucalyptus/3/wb
[6] eucalyptus/3/bare
[7] dogwood/3/fun
[8] dogwood/3/bare
```

### DIFF
--- a/bin/activate
+++ b/bin/activate
@@ -23,6 +23,7 @@ function select_release(){
     read -r -a releases <<< $(
         find "${PROJECT_DIRECTORY}/releases/" -maxdepth 4 -name Dockerfile |
         sed "s|${PROJECT_DIRECTORY}/releases/\\(.*\\)/Dockerfile|\\1|g" |
+        sort -fir |
         xargs
     )
     n_releases=${#releases[@]}


sampaccoud (approved):
Nice!

madmatah (changes requested):
Nice, thanks!

Just a comment about the commit message subject. As described in our [gitbook](https://openfun.gitbooks.io/handbook/content/git.html), you should add a scope, i.e. : 
`🎨 (project) sort release in reverse order for bin/activate`

slandrault:
> As described in our [gitbook](https://openfun.gitbooks.io/handbook/content/git.html), you should add a scope, i.e. :
> `🎨 (project) sort release in reverse order for bin/activate`
True. What kind of project would be most accurate here? "(activate)"?

madmatah:
> > As described in our [gitbook](https://openfun.gitbooks.io/handbook/content/git.html), you should add a scope, i.e. :
> > `🎨 (project) sort release in reverse order for bin/activate`
> 
> True. What kind of project would be most accurate here? "(activate)"?

I looked into the commit history and the scope `(bin)` is sometime used for the scripts in the `bin/` directory.
Also, the `(project)` scope is often used for all that concerns the project structure / build tools.

madmatah (approved):

jmaupetit (commented):
Why not sorting releases alphabetically?

slandrault:
> Why not sorting releases alphabetically instead?

To get the most recent first (as we are usually reading from top to bottom) and therefore:
- Easily get the history of releases
- having only one digit to press instead of two when number of versions increases (and thus less prone to typing error)

bin/activate
```diff
@@ -23,6 +23,7 @@ function select_release(){
     read -r -a releases <<< $(
         find "${PROJECT_DIRECTORY}/releases/" -maxdepth 4 -name Dockerfile |
         sed "s|${PROJECT_DIRECTORY}/releases/\\(.*\\)/Dockerfile|\\1|g" |
+        sort -f -i -r |
```
jmaupetit, line 26: Are `-f` and `-i` flags required in this case?
```suggestion
        sort -r |
```

jmaupetit (commented):
I just realize that you are talking of OpenEdx releases, not `openedx-docker`... and was looking for black magic I don't understand in your command :sweat_smile:

bin/activate
slandrault: There are not *required* but more for "futureproofing" the script :D . No problem to remove then to make the script maybe more understandable.

jmaupetit: Do as you please :hugs: I you leave all flags, can you merge them (`sort -fir`)?

jmaupetit (approved):

PR #241, opened 2020-09-01
🚸(bin) autoexport variables if source when activating

slandrault:
## Purpose

Make our life easier.

## Proposal

To make our life easier, the `bin/activate` command will now auto-export the environment variable if started with a leading `.` or `source` operator like so: `. bin/activate`.

If the leading dot (or source) is omitted, then we fallback to regular behavior, printing the variables export commands to run.

(code inspired from: https://github.com/openfun/richie-site-factory/blob/master/bin/activate#L44)

### DIFF
--- a/bin/activate
+++ b/bin/activate
@@ -55,6 +55,7 @@ function select_release(){
 function activate_release(){
 
     declare release="$1"
+    declare -i is_sourced="$2"
 
     release_path="${RELEASES_ROOT}/${release}"
     env_file_path="${release_path}/${ENV_FILENAME}"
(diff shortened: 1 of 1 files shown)


bin/activate
```diff
@@ -69,13 +69,19 @@ function activate_release(){
         exit 21
     fi
 
-    echo -e "\\n# Copy/paste ${release} environment:"
-    cat "${env_file_path}"
-
-    echo -e "\\n# Or run the following command:"
-    echo -e ". ${env_file_path}"
+    if [[ "${BASH_SOURCE[0]}" != "${0}" ]] ; then
```
jmaupetit, line 72: I've read that the recommended way to do so is to use the following instead:

```suggestion
# Check if this script is being sourced or executed. Explanation: Bash allows
# return statements only from functions and, in a script's top-level scope, only
# if the script is sourced.
(return 2> /dev/null) && sourced=1 || sourced=0

if [[ ${sourced} == 0 ]]; then
```

jmaupetit (commented):
Looks like it's not working with `zsh` (at least with my current configuration): it kills the parent shell when sourced!

Maybe we can still accept this bash-users feature as most people are using this shell (this is a strong assumption :sweat_smile:).

bin/activate
```diff
@@ -55,6 +55,7 @@ function select_release(){
 function activate_release(){
 
     declare release="$1"
+    declare is_sourced="$2"
```
jmaupetit, line 58: Make it an integer!

```suggestion
    declare -i is_sourced="$2"
```

bin/activate
```diff
@@ -69,15 +70,22 @@ function activate_release(){
         exit 21
     fi
 
-    echo -e "\\n# Copy/paste ${release} environment:"
-    cat "${env_file_path}"
+    if [[ ${is_sourced} != 0 ]]; then
```
jmaupetit, line 73: Maybe the following is more explicit?

```suggestion
    if [[ ${is_sourced} -eq 1 ]]; then
```

slandrault: Thanks for the tips!

jmaupetit (approved):

Issue #244, opened 2020-09-02
Speed-up CI by building only what has changed


sampaccoud:
## Expected Behavior

CI builds should be kept as fast as possible.

## Actual Behavior

The CI builds all releases all the time. Now that we are adding 2 releases for ironwood, it will slow it down even more.

## Proposal

We could generate the CI configuration file, on the basis of what has changed, in order to rebuild only the impacted releases : 
- If what has changed is in the directory of a release : rebuild only this release
- If what has changed impacts all releases (e.g. in the docker directory? anything else?) : rebuild all releases

Something similar was done in [richie site factory](https://github.com/openfun/richie-site-factory). See the "[bin/ci update](https://github.com/openfun/richie-site-factory/blob/master/bin/ci#L70)" command and CircleCI's documentation about [packing a configuration](https://circleci.com/docs/2.0/local-cli/#packing-a-config) with the CLI.

PR #245, opened 2020-09-04
Lightning the load on circle-ci with selective release building

slandrault:
## Purpose

As mentionned in #244 build every release every time.

The aim of this development is to lighten the load on circle-ci by building only project that have changed. This way, no more extra long build will be made and this repo will require less resources from circle-ci.

## Proposal

By watching changes in `releases/` folder, the `bin/ci update` command will generate a config.yml file properly configured for building only useful projects. This also means that from now on, the `bin/ci update` will have to be run before each commit to make sure that the `config.yml` file is up to date (if not a job will check it and block release building anyway)

Also, to have a consistent naming in the config file, master-bare project have been moved to master.0-bare with the associated folder structure.


bin/ci
```diff
@@ -76,8 +79,89 @@ function checkpoint(){
 
 # Get a list of changed files in the current branch
 function get_changes() {
+  git whatchanged --name-only --pretty="" "origin/master..HEAD" | sort -u
+}
+
+
+function check_changes() {
+  # to="${CIRCLE_SHA1:-HEAD}"
+  # git whatchanged --name-only --pretty="" origin/master.."${to}" | grep "${1}" &> /dev/null
```
jmaupetit, line 90: I can't remember why this `CIRCLE_SHA1` to `HEAD` fallback was implemented.

```suggestion
```

bin/ci
```diff
@@ -130,9 +132,12 @@ function update_sources() {
     done <  <(find "releases" -maxdepth 3 -mindepth 3 -type d -print0 | sort -z)
 
     # Replace in config.tpl
-    awk '/{{ WORKFLOW_JOBS_LIST }}/{system("cat .circleci/src/workflows/edxapp/@jobs.yml");next}1' .circleci/src/config.tpl > .circleci/config.yml.part
-    awk '/{{ JOBS_LIST }}/{system("cat .circleci/src/jobs/@jobs.yml");next}1' .circleci/config.yml.part > .circleci/config.yml
-    rm .circleci/config.yml.part
+    awk '/{{ WORKFLOW_JOBS_LIST }}/{system("cat .circleci/src/workflows/edxapp/@jobs.yml");next}1' .circleci/src/config.tpl > .circleci/config.yml.part1
+    awk '/{{ JOBS_LIST }}/{system("cat .circleci/src/jobs/@jobs.yml");next}1' .circleci/config.yml.part1 > .circleci/config.yml.part2
```
jmaupetit, line 136: Can't you use `sed` instead?

bin/ci
slandrault: Oh yes, for my test I changed for "HEAD" to test locally. I should probably revert to the `${CIRCLE_SHA1:-HEAD}` as it was before no?

jmaupetit (approved):

madmatah (approved):

sampaccoud (approved):
LGTM

Issue #204, opened 2020-03-10
Improve user experience when trying to build/run the project with a root user


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
The `make bootstrap` command fails when launched via a root user or with sudo. We need to fail early, nicely and explain what is the problem of running Docker with a root user instead of failing with a difficult error after screwing up the user's file system.

See this discussion for details: https://github.com/openfun/openedx-docker/issues/203

**Describe the solution you'd like**
Any make command should start by checking that the user does not have the `0` id with:
```bash
id -u
```

**Describe alternatives you've considered**
An alternative could be to modify the Dockerfile to allow building it with a root user.
In this case we can still display a warning but let the user continue if s.he knows what s.he is doing.

jmaupetit:
I would prefer the alternative solution and not enforcing to run commands without using the root user (or sudo). The warning message should be explicit enough so that our users are aware of the potential risk they encounter.

PR #247, opened 2020-09-08
🔨(project) warn user when running make bootstrap from root user

slandrault:
## Purpose

In order to not mess up user filesystem, the build should not be run as a root user.
This commit add a `check-root-user` make target that will check if the user running the make is not with an id = 0 (root).

## Proposal

To guard any make target from being run as root, a `check-root-user` target can be called first. If the user is root, then the command will fail with a warning message. The warning/fail behavior can be override by setting a `ALLOW_ROOT` variable like: 
```
$ sudo ALLOW_ROOT=1 make bootstrap
```

Fixes https://github.com/openfun/openedx-docker/issues/204

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,15 +159,33 @@ jobs:
   #
   # Note that the job name should match the EDX_RELEASE value
 
-  # No changes detected for dogwood.3-bare
-  # No changes detected for dogwood.3-fun
-  # No changes detected for eucalyptus.3-bare
-  # No changes detected for eucalyptus.3-wb
-  # No changes detected for hawthorn.1-bare
-  # No changes detected for hawthorn.1-oee
-  # No changes detected for ironwood.2-bare
-  # No changes detected for ironwood.2-oee
-  # No changes detected for master.0-bare
+  # Run jobs for the dogwood.3-bare release
+  dogwood.3-bare:
+    <<: [*defaults, *build_steps]
+  # Run jobs for the dogwood.3-fun release
+  dogwood.3-fun:
+    <<: [*defaults, *build_steps]
+  # Run jobs for the eucalyptus.3-bare release
+  eucalyptus.3-bare:
+    <<: [*defaults, *build_steps]
+  # Run jobs for the eucalyptus.3-wb release
+  eucalyptus.3-wb:
+    <<: [*defaults, *build_steps]
+  # Run jobs for the hawthorn.1-bare release
+  hawthorn.1-bare:
+    <<: [*defaults, *build_steps]
+  # Run jobs for the hawthorn.1-oee release
+  hawthorn.1-oee:
+    <<: [*defaults, *build_steps]
+  # Run jobs for the ironwood.2-bare release
+  ironwood.2-bare:
+    <<: [*defaults, *build_steps]
+  # Run jobs for the ironwood.2-oee release
+  ironwood.2-oee:
+    <<: [*defaults, *build_steps]
+  # Run jobs for the master.0-bare release
+  master.0-bare:
+    <<: [*defaults, *build_steps]
 
   # Hub job
   hub:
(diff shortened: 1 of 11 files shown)


slandrault:
For the moment, the solution is implemented as an "fail if run as root". I'll evaluate the feasibility of achieving similar feature with a warning+confirmation instead of failing all the time.

slandrault:
The target has been updated. In case of root call, the command will now ask for confirmation before continuation instead of leaving.

releases/dogwood/3/fun/Dockerfile
```diff
@@ -215,12 +215,8 @@ RUN apt-get update && \
...
 RUN groupadd --gid ${DOCKER_GID} edx || \
       echo "Group with ID ${DOCKER_GID} already exists." && \
-    useradd \
-      --create-home \
-      --home-dir /home/edx \
-      --uid ${DOCKER_UID} \
-      --gid ${DOCKER_GID} \
-      edx && \
+    useradd --create-home --home-dir /home/edx --uid ${DOCKER_UID} --gid ${DOCKER_GID} edx || \
+      echo "User with ID ${DOCKER_UID} already exists." && \
```
jmaupetit, line 219: We are not sure of the reason why the useradd command fails:

```suggestion
      echo "Skip user creation (user with ID ${DOCKER_UID} already exists?)" && \
```

slandrault: Good point

jmaupetit (commented):
Maybe other commands that create/access mounted volumes should also have the `ensure-non-root` dependency (_e.g._ `build`, `build-dev`, `tree`), WDYT?

slandrault:
> Maybe other commands that create/access mounted volumes should also have the `ensure-non-root` dependency (_e.g._ `build`, `build-dev`, `tree`), WDYT?

They could, but that means that when bootstrap will be run it will ask for confirmation when reaching each of those target... Maybe not very user friendly?

jmaupetit:
> Maybe not very user friendly?

At least the user will understand that this is not the recommended way to do things :joy:

slandrault:
A side effect would be that the bootstrap won't be able to run by itself until completion because the user would have to validate several times that he is ok to continue as sudo...

sampaccoud:
We could request to set an env variable to be allowed to run any command as sudo?

slandrault:
> We could request to set an env variable to be allowed to run any command as sudo?

:+1:  Sounds like a an elegant solution. If the variable is set then go, if not then warning message and fail nicely.

Makefile
```diff
@@ -342,6 +357,19 @@ tree: \
 tree:  ## create data directories mounted as volumes
 .PHONY: tree
 
+verify-root:  # Make sure the user calling this is not currently root
```
lunika, line 360: In your commit message you called it `ensure-not-root`

lunika, line 360:
```suggestion
verify-root:  ## Make sure the user calling this is not currently root
```

Makefile
```diff
@@ -166,14 +168,19 @@ create-symlinks:  ## create symlinks to local configuration (mounted via a volum
 	  ln -sf /config/lms/root_urls.py /edx/app/edxapp/edx-platform/lms/"
 .PHONY: create-symlinks
 
-demo-course: fetch-demo  ## import demo course from edX repository
+# Import demo course from edX repository
```
lunika, line 171:
```suggestion
```

Makefile
```diff
@@ -166,14 +168,19 @@ create-symlinks:  ## create symlinks to local configuration (mounted via a volum
...
+demo-course: \
+	verify-root \
+	fetch-demo
+demo-course:
```
lunika, line 175:
```suggestion
demo-course: ## import demo course from edX repository
```

Makefile
```diff
@@ -231,7 +240,10 @@ dev-ui-toolkit:
 	  bash -c "cd node_modules/edx-ui-toolkit || exit 0 && npm install"
 .PHONY: dev-ui-toolkit
 
-dev-watch: tree  ## start assets watcher (front-end development)
+# Start assets watcher (front-end development)
```
lunika, line 243:
```suggestion

```

Makefile
```diff
@@ -231,7 +240,10 @@ dev-ui-toolkit:
...
+dev-watch: \
+	verify-root \
+	tree \
```
lunika, line 246:
```suggestion
	tree
dev-watch: ## Start assets watcher (front-end development)
```

releases/dogwood/3/bare/Dockerfile
```diff
@@ -223,12 +223,8 @@ RUN apt-get update && \
 
 RUN groupadd --gid ${DOCKER_GID} edx || \
       echo "Group with ID ${DOCKER_GID} already exists." && \
-    useradd \
-      --create-home \
-      --home-dir /home/edx \
-      --uid ${DOCKER_UID} \
-      --gid ${DOCKER_GID} \
-      edx && \
+    useradd --create-home --home-dir /home/edx --uid ${DOCKER_UID} --gid ${DOCKER_GID} edx || \
```
lunika, line 226: You should keep the multiline command, we use this format to improve readability

.circleci/config.yml
```diff
@@ -144,17 +144,33 @@ jobs:
   #
   # Note that the job name should match the EDX_RELEASE value
```
lunika, line 145: in your commit title: 

`s/prevent make bootstrap to be run from root user/prevent make bootstrap to be run as root user/`

lunika (commented):
The makefile is self documented with the `help` task. All comments starting by `##` after a task name must be kept and you will see them when you run `make` or `make help`

Makefile
jmaupetit: What about `check-root-user`?

Makefile
slandrault: If we do that then we break the "codestyle" of the file. Everytime a target is on multiple line then the comment is above the command, only when there is no / the comment is next to the target name.

lunika: Take a look at the `help` task, this task is using comments starting with `##`

lunika: Also look this article https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html

slandrault: Ohhhh ok yes I understand why now! Thx!

Makefile
```diff
@@ -145,6 +147,18 @@ build:  ## build the edxapp production image
 	$(COMPOSE) build nginx
 .PHONY: build
 
+check-root-user:  ## Make sure the user calling this is not currently root
+	@if [ $(shell id -u) -eq 0 ]; \
```
jmaupetit, line 151: It is generally safer to use the double bracket syntax:

```suggestion
	@if [[ $(shell id -u) -eq 0 ]]; \
```

Reference: http://mywiki.wooledge.org/BashGuide/Practices#Bash_Tests

jmaupetit (changes requested):
Rules that may create files and thus should have the `check-root-user` dependency are:

- `build`
- `clean`
- `create-symlinks`
- `demo-course`
- `dev-assets`
- `dev-build`
- `dev-install`
- `dev-watch`
- `fetch-demo`
- `fetch-release`
- `migrate`
- `run`
- `tree`

Makefile
```diff
@@ -114,6 +114,7 @@ $(FLAVORED_EDX_RELEASE_PATH)/src/edx-demo-course/README.md:
 	tar xzf /tmp/edx-demo.tgz -C $(FLAVORED_EDX_RELEASE_PATH)/src/edx-demo-course --strip-components=1
 
 bootstrap: \
+  check-root-user \
```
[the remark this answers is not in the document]
slandrault: Fair point.

releases/dogwood/3/bare/Dockerfile
```diff
@@ -222,13 +222,14 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/*
```
jmaupetit, line 222: Your commit message scope should be: `flavors/*` and I would rephrase your commit title as:

:hammer:(flavors/*) allow builds using the root user

If you hesitate on a scope, check the project's commits history!

jmaupetit (approved):

PR #248, opened 2020-09-08
🐛(dogwood/3/fun) pin `django-classy-tags` version to 0.8.0

madmatah:
## Purpose

Recent builds of dogwood.3-fun install `django-classy-tags` 2.0.0, that is not compatible anymore with python 2.x.

Same problem with eucalyptus flavor.

## Proposal

- [x] (eucalyptus + dogwood) pin `django-classy-tags` version to 0.8.0
- [x] bump to dogwood.3-fun-1.14.1
- [x] bump to eucalyptus.3-wb-1.9.3

### DIFF
--- a/releases/dogwood/3/fun/CHANGELOG.md
+++ b/releases/dogwood/3/fun/CHANGELOG.md
@@ -9,6 +9,8 @@ release.
 
 ## [Unreleased]
 
+## [dogwood.3-fun-1.14.1] - 2020-09-08
+
 ### Fixed
 
 - Pin django-classy-tags to 0.8.0 to avoid breaking changes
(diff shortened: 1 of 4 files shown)


lunika (approved):

sampaccoud:
@madmatah I think the other flavors will have the same problem !?

sampaccoud:
So should we pin it to 0.8 instead if that's what Hawthorn is running on!?

madmatah:
> So should we pin it to 0.8 instead if that's what Hawthorn is running on!?

@sampaccoud  Yes we can do that to be sure.
But I don't see any impacting change in their [Changelog](https://github.com/django-cms/django-classy-tags/blob/master/CHANGELOG.rst) between 0.8.0 and 1.0.0

jmaupetit (approved):

madmatah:
Done!

I'm ready for a final review if the CI checks pass :slightly_smiling_face:

sampaccoud:
@madmatah are the `bare` flavors not concerned?

madmatah:
> @madmatah are the `bare` flavors not concerned?

You're right, `dogwood/3/bare` and `eucalyptus/3/bare` are also affected by this problem.
But in order to fix it, a change needs to be made in the bare flavors to be able to add custom dependencies.
I created #249 to deal with this problem in another PR.

sampaccoud (approved):

PR #253, opened 2020-09-21
➕(ironwood/2/oee) install scormxblock-xblock package

lunika:
## Purpose

We install scorm xblock to allow instructor to add scorm content in a
course. We must install unzip, it is a dependency of this package.
scormxblock-xblock package is host on our gemfury repo, we want to pin
an exact version and not installing it via github.

## Proposal

- [x] install scormxblock-xblock package
- [x] fix superuser makefile task targeting wrong service

### DIFF
--- a/env.d/development
+++ b/env.d/development
@@ -26,3 +26,7 @@ OAUTH_OIDC_ISSUER=http://edx:8073/oauth2
 
 # API
 EDX_API_KEY=FakeEdXAPIKey
+
+HTTPS=off
+LMS_BASE=localhost:8073
+CMS_BASE=localhost:8083
--- a/releases/ironwood/2/oee/requirements.txt
+++ b/releases/ironwood/2/oee/requirements.txt
@@ -5,6 +5,7 @@ fonzie==0.2.1
 
 # ==== xblocks ====
 configurable_lti_consumer-xblock==1.2.3
+openedx-scorm-xblock==9.2.0
 
 # ==== third-party apps ====
 celery-redis-sentinel==0.3.0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,15 +144,33 @@ jobs:
   #
   # Note that the job name should match the EDX_RELEASE value
 
-  # No changes detected for dogwood.3-bare
-  # No changes detected for dogwood.3-fun
-  # No changes detected for eucalyptus.3-bare
-  # No changes detected for eucalyptus.3-wb
-  # No changes detected for hawthorn.1-bare
-  # No changes detected for hawthorn.1-oee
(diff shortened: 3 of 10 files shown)


env.d/development
```diff
@@ -26,3 +26,7 @@ OAUTH_OIDC_ISSUER=http://edx:8073/oauth2
 
 # API
 EDX_API_KEY=FakeEdXAPIKey
+
+HTTPS=off
+LMS_BASE=localhost:8073
+CMS_BASE=localhost:8083
```
sampaccoud, line 32: Why not include the removal of *_BASE default values from docker_run_production in this PR?

lunika: You are right

lunika: Done

sampaccoud (approved):

releases/ironwood/2/oee/requirements.txt
```diff
@@ -1,10 +1,12 @@
 # Open edX extended: dependencies
+--extra-index-url https://pypi.fury.io/openfun/
 
 # ==== core ====
 fonzie==0.2.1
 
 # ==== xblocks ====
 configurable_lti_consumer-xblock==1.2.3
+scormxblock-xblock==1.0.3
```
regisb, line 9: May I suggest instead https://pypi.org/project/openedx-scorm-xblock/ ? This fork includes a few important improvements, including code cleaning, no more dependency on unzip and object storage support.

lunika: I don't see the point of `object storage support`. For what I understood you just allow to change the directory inside the `/media` repository. Instead of storing it in `/media/scorm` we can save them in `/media/whatever`. Am I wrong ? Did I miss something ?

regisb: Object storage allows you to store data in S3 or, in my case, MinIO. This is useful when you want to share media files between different servers/pods.

Also, the xblock that I developed is properly maintained and already compatible with Juniper.

Just my 2 cents -- I don't get paid when my fork is installed :) I just feel like it's a shame to install an xblock that is no longer maintained.

lunika: > Object storage allows you to store data in S3 or, in my case, MinIO. This is useful when you want to share media files between different servers/pods.

I thought you have a special support on your package in fact, but it's not the case, this is why I ask you if I miss something.

regisb: The original repo has a weird mix of `default_storage` and `shutil.rmtree`/`os.path`, so it is not really compatible with object storage.

lunika:
I close this PR, this xblock will not be added in our default image.

PR #272, opened 2021-01-04
👷(project) allow Docker Hub anonynous mode

jmaupetit:
## Purpose

We need to allow external contributions to run the CI, but without injecting secrets in the environment for security concerns.

## Proposal

When the $DOCKER_USER environment variable is not set, we allow to pull Docker images anonymously from Docker Hub.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,10 +23,14 @@ build_steps: &build_steps
     - checkout
 
     # Login to DockerHub with encrypted credentials stored as secret
-    # environment variables (set in CircleCI project settings)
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -168,6 +172,9 @@ jobs:
   # No changes detected for ironwood.2-bare
   # No changes detected for ironwood.2-oee
   # No changes detected for master.0-bare
+  # No changes detected for master.bare-config
```
sampaccoud, line 175: be careful, you must delete your folder "bare" in "master"

sampaccoud (approved):

PR #273, opened 2021-01-06
Upgrade fun-apps to v5.7.0 and add settings for course run synchronization with richie

sampaccoud:
## Purpose

In [fun-apps](https://github.com/openfun/fun-apps) v5.7.0 was added the possibility to call a web hook each time a course is published, in order to synchronize its dates with a remote catalog like richie.

## Proposal

- [x] Bump fun-apps to v5.7.0 in Python dependency requirements,
- [x] Release edxapp `dogwood.3-fun-1.18.0`

I had to update the development stack to work on this feature:
- [x] Improve the docker-compose configuration to allow calling richie server-to-server from the cms,
- [x] Add configurable `COURSE_HOOK` setting with a default value that connects with the [richie development stack](https://github.com/openfun/richie),

### DIFF
--- a/env.d/development
+++ b/env.d/development
@@ -26,3 +26,6 @@ OAUTH_OIDC_ISSUER=http://edx:8073/oauth2
 
 # API
 EDX_API_KEY=FakeEdXAPIKey
+
+# Richie synchronization
+COURSE_HOOKS=[{"url": "https://richie:8070/api/v1.0/course-runs-sync/", "secret": "shared secret", "verify": false}]
--- a/releases/dogwood/3/fun/CHANGELOG.md
+++ b/releases/dogwood/3/fun/CHANGELOG.md
@@ -9,13 +9,15 @@ release.
 
 ## [Unreleased]
 
+## [dogwood.3-fun-1.18.0] - 2021-01-06
+
 ### Added
 
 - Upgrade fun-apps to v5.7.0 to allow synchronizing course runs with richie
 
 ### Fixed
 
-- Force Django old-release de-installation due to recent changes in the new
+- Force Django old-release uninstallation due to recent changes in the new
   pip dependencies resolver
 
 ## [dogwood.3-fun-1.17.0] - 2020-12-10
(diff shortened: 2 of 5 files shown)


releases/dogwood/3/fun/CHANGELOG.md
```diff
@@ -9,9 +9,16 @@ release.
...
+## [dogwood.3-fun-1.18.0] - 2021-01-06
+
+### Added
+
+- Upgrade fun-apps to v5.7.0 to allow synchronizing course runs with richie
+
 ### Fixed
 
-- Force Django old-release de-installation due to recent changes in the new pip dependencies resolver
+- Force Django old-release de-installation due to recent changes in the new
```
jbpenrath, line 20:
```suggestion
- Force Django old-release uninstallation due to recent changes in the new
```

env.d/development
```diff
@@ -26,3 +26,6 @@ OAUTH_OIDC_ISSUER=http://edx:8073/oauth2
 
 # API
 EDX_API_KEY=FakeEdXAPIKey
+
+# Richie synchronization
+COURSE_HOOKS=[{"url": "https://richie:8070/api/v1.0/course-runs-sync/", "secret": "shared secret", "verify": false}]
```
jbpenrath, line 31: `richie` name will be only resolved in the production enviromnent.
It is relevant to have this information into `development` to see this settings with an example but I think it will not work in a development environment.

jbpenrath (approved):

env.d/development
sampaccoud: Or should we add the "outside" network in the development container as well to make it work? I'm sure I'm missing something :stuck_out_tongue:

jbpenrath: Hmm the url will be still wrong as `richie` service name is defined in Matsuo right? To make it works in develpment, we should use the local richie app container name.

sampaccoud: Oh! ok I get your point. I forgot to push a part of my work on the richie project that declares the `richie` host: https://github.com/openfun/richie/pull/1236

releases/dogwood/3/fun/CHANGELOG.md
sampaccoud: Makes sense :+1:

PR #283, opened 2021-04-06
🐛(all) fix get-pip url

sampaccoud:
## Purpose

The get-pip script has moved to a new address.

## Proposal

Replace the url in each DockerFile.


sampaccoud:
Unsuccessful attempt to fix all flavors build at once.

UbuntuEvangelist:
I am using Ubuntu 20.04 and facing the same problem. 

Screenshot from 2021-11-08 09-27-37

PR #284, opened 2021-04-08
♻️(dogwood/3/fun) refactor build strategy

jmaupetit:
## Purpose

Due to python 2 compatibility, some dependencies can no longer be installed for the dogwood release, breaking our Docker image build workflow.

## Proposal

To mitigate the issue, we decided to update our Docker image build strategy to start from the last-usable build, update fun-specific dependencies and compile/collect static files in a new image.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,17 +164,15 @@ jobs:
   # Note that the job name should match the EDX_RELEASE value
 
   # No changes detected for dogwood.3-bare
-  # No changes detected for dogwood.3-fun
+  # Run jobs for the dogwood.3-fun release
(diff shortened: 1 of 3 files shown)


jmaupetit:
Should I add the legacy `Dockerfile` (as `Dockerfile.legacy`) to easily track how was built the latest working image?

jbpenrath (approved):
Looks good to me but a second aproval could be fine !

About your question to old Dockerfile in a `Dockerfile.legacy`, I think that using git to retrieve the old version of this file should be enough.

lunika (approved):
Same as @jbpenrath the git history is here for that.

PR #287, opened 2021-04-13
Fix dogwood.3-fun build

jmaupetit:
## Purpose

Latest dogwood.3-fun build does not reflect expected changes (outdated dependencies).

## Proposal

The `production` stage should copy build dependencies, static files and sources from the `builder` stage.


lunika (approved):

jbpenrath (approved):
🚢

PR #291, opened 2021-05-31
Fix eucalyptus build by upgrading setuptools

sampaccoud:
## Purpose

We previously upgraded `pip` to fix installing the github dependencies but it broke again. In fact, it's `setuptools`  that must be upgraded.

## Proposal

Replace the `pip` upgrade by a `setuptools` upgrade.


jmaupetit (approved):
In your commit message `s/upgrade/upgrading/`?

sampaccoud:
> In your commit message `s/upgrade/upgrading/`?

Well spotted ;-) 
Done.

PR #292, opened 2021-06-01
Fix/ironwood 2 build

jbpenrath:
## Purpose

Currently it is not possible to build OpenEdX Ironwood. There are 2 causes:
- The link to download pip install for python 2.7 has changed
- This OpenEdX release used py2neo 3.1.2 but this dependency is no more available from pypi.

## Proposal

We have first to update the link to download pip install for python 2.7. Then as we cannot install py2neo==3.1.2 from pypi, we have to patch requirement files to download this dependency version directly from its github repository.

- [x] Fix ironwood/2/bare build
- [x] Fix ironwood/2/oee build

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,16 +165,16 @@ jobs:
 
   # No changes detected for dogwood.3-bare
   # No changes detected for dogwood.3-fun
-  # Run jobs for the eucalyptus.3-bare release
-  eucalyptus.3-bare:
-    <<: [*defaults, *build_steps]
-  # Run jobs for the eucalyptus.3-wb release
-  eucalyptus.3-wb:
-    <<: [*defaults, *build_steps]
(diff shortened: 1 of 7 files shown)


sampaccoud:
Actually I'm wondering if it wouldn't work to just pre-installed pyneo like this in the Dockerfile. But the advantage of what you did is that it will be easy to further modify the requirements the next time the build breaks 😅

sampaccoud (approved):
Maybe add a CHANGELOG then GTM!

jbpenrath:
> Actually I'm wondering if it wouldn't work to just pre-installed pyneo like this in the Dockerfile. But the advantage of what you did is that it will be easy to further modify the requirements the next time the build breaks 😅

Yep. First I ran a command to search & replace py2neo into requirements files but I thought to have a separate patch file will be more explicit and as you said if later we need to install/replace/pin a dependency we have now a file to do this.

PR #300, opened 2021-09-28
🔒️(all) set session cookie secure to True as default

sampaccoud:
## Purpose

We relied on OpenEdX's default value for the `SESSION_COOKIE_SECURE` setting but it is unsecure.

## Proposal

Directly use `True` as default value for the `SESSION_COOKIE_SECURE` setting.


jmaupetit (approved):

sampaccoud:
@jmaupetit dogwood.3.bare won't build anymore so I proposed to remove it...

jmaupetit:
> @jmaupetit dogwood.3.bare won't build anymore so I proposed to remove it...

Do it! :fire:

PR #303, opened 2022-01-26
♻️(dogwood/3/fun) revert to building the image from scratch

sampaccoud:
### Purpose

Since pypi banned clients that don't support SNI, we did not manage to build the image from scratch and tried to build each image on top of the previous image.

This strategy causes our images to inflate and is not sustainable.

### Proposal

We now propose to revert building the image from a plain  Ubuntu 12.04 image and upgrade its Python 2 to 2.7.18, the last version available which supports SNI.

Following @jbpenrath's suggestion to fix a typo, all images had to rebuild so I also added a fix of the Eucalpytus images.

### DIFF
--- a/releases/dogwood/3/fun/Dockerfile
+++ b/releases/dogwood/3/fun/Dockerfile
@@ -236,7 +236,7 @@ RUN groupadd --gid ${DOCKER_GID} edx || \
     git config --global user.name edx && \
     git config --global user.email edx@example.com
 
-# To prevent permission issues related to the non-priviledged user running in
+# To prevent permission issues related to the non-privileged user running in
 # development, we will install development dependencies in a python virtual
 # environment belonging to that user
 RUN pip install virtualenv==16.7.9
@@ -261,7 +261,7 @@ RUN chown -R ${DOCKER_UID}:${DOCKER_GID} /tmp/sass-cache
 # issues with volumes (host folders)
 USER ${DOCKER_UID}:${DOCKER_GID}
 
-# Create the virtualenv with a non-priviledged user
+# Create the virtualenv with a non-privileged user
 RUN virtualenv -p python2.7 --system-site-packages /edx/app/edxapp/venv
 
 # Install development dependencies in a virtualenv (we need to downgrade pip
--- a/releases/dogwood/3/fun/Dockerfile
+++ b/releases/dogwood/3/fun/Dockerfile
@@ -1,32 +1,149 @@
 # EDX-PLATFORM multi-stage docker build
 
+# Change release to build, by providing the EDX_RELEASE_REF build argument to
+# your build command:
+#
+# $ docker build \
+#     --build-arg EDX_RELEASE_REF="named-release/dogwood.3" \
+#     -t edxapp:dogwood.3-fun \
+#     .
 ARG DOCKER_UID=1000
 ARG DOCKER_GID=1000
-ARG EDX_BASE_IMAGE=fundocker/edxapp:dogwood.3-fun-2.1.1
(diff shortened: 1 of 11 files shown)


releases/dogwood/3/fun/Dockerfile
```diff
@@ -1,45 +1,188 @@
...
+ARG EDX_ARCHIVE_URL=https://github.com/openfun/edx-platform/archive/dogwood.3-fun-5.3.3.tar.gz
 ARG EDXAPP_STATIC_ROOT=/edx/app/edxapp/staticfiles
 ARG NGINX_IMAGE_NAME=fundocker/openshift-nginx
 ARG NGINX_IMAGE_TAG=1.13
+ARG PYTHON_VERSION=2.7.18
+
 
 # === BASE ===
-FROM ${EDX_BASE_IMAGE} as base
+FROM ubuntu:12.04 as base
```
jmaupetit, line 21: Why not updating the `EDX_BASE_IMAGE` default value instead?

jmaupetit (commented):
Python and pip are both installed twice. Not sure which version is running finally. I have the feeling that it requires an advised clean up  :thinking:

releases/dogwood/3/fun/Dockerfile
sampaccoud: That's because I started fresh from the Dockerfile before the refactoring.
I think we should keep it like this because:
- this is not something you want to change, never ever OMG :sweat_smile: 
- images for all other releases are like this

sampaccoud:
> Python and pip are both installed twice. Not sure which version is running finally. I have the feeling that it requires an advised clean up thinking

:thinking: I don't know if this is a problem and I'm not sure there's a clean way to make this kind of surgery ? This would not make the image smaller so is there a point? The image is back to a reasonable size (divided by 3 compared to what it would have been building this new version on the previous one) and further versions will keep a stable size:
```
REPOSITORY                                      TAG                   IMAGE ID       CREATED         SIZE
edxapp                                          dogwood.3-fun         1199e66af05b   7 hours ago     1.6GB
```

About mixing up versions, the application seems to work perfectly and  calling "python" or "pip" seems to do what we expect:
```bash
$ pip --version
pip 20.3.4 from /usr/local/lib/python2.7/site-packages/pip (python 2.7)
$ python --version
Python 2.7.18
```

jmaupetit (approved):

releases/dogwood/3/fun/Dockerfile
```diff
@@ -95,10 +236,23 @@ RUN groupadd --gid ${DOCKER_GID} edx || \
     git config --global user.name edx && \
     git config --global user.email edx@example.com
 
+# To prevent permission issues related to the non-priviledged user running in
```
jbpenrath, line 239:
```suggestion
# To prevent permission issues related to the non-privileged user running in
```

releases/dogwood/3/fun/Dockerfile
```diff
@@ -107,17 +261,39 @@ RUN chown -R ${DOCKER_UID}:${DOCKER_GID} /tmp/sass-cache
 # issues with volumes (host folders)
 USER ${DOCKER_UID}:${DOCKER_GID}
 
+# Create the virtualenv with a non-priviledged user
```
jbpenrath, line 264:
```suggestion
# Create the virtualenv with a non-privileged user
```

jbpenrath (approved):
LGTM

releases/dogwood/3/fun/Dockerfile
sampaccoud: Found on other other flavors as well :+1:

PR #304, opened 2022-01-26
♻️(all) use Nginx Inc's unprivileged image instead of our own

sampaccoud:
### Purpose

We were building our own specific image able to run as a non-root user. Since Nginx Inc is building an official unprivileged image, we should use it.

## Proposal

Use the following default definition for the nginx image:
- ARG NGINX_IMAGE_NAME=nginxinc/nginx-unprivileged
- ARG NGINX_IMAGE_TAG=1.20

### DIFF
--- a/releases/dogwood/3/fun/CHANGELOG.md
+++ b/releases/dogwood/3/fun/CHANGELOG.md
@@ -11,6 +11,7 @@ release.
 
 ### Changed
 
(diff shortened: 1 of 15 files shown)


jmaupetit (approved):

sampaccoud:
@jmaupetit sorry the first attempt didn't build. I had to add a USER switch before copying the static files. You might want to take a look at it.

jmaupetit (approved):

PR #307, opened 2022-04-08
Dogwood.3 fun/install fonzie

jbpenrath:
## Purpose

To enable Joanie within Richie, we need that the authentication endpoint returns a JWT token. Currently, this task is managed by Fonzie so we need to setup fonzie within `dogwood.3-fun`

## Proposal

- [x] Install fonzie 0.3.0 and restframework.simple-jwt
- [x] Update settings to append fonzie api routes and configure simple jwt


releases/dogwood/3/fun/config/lms/docker_run_production.py
```diff
@@ -1222,7 +1230,8 @@
     "videoproviders",
 )
 
-ROOT_URLCONF = "fun.lms.urls"
+# Override edX LMS urls with ours
+ROOT_URLCONF = "lms.root_urls"
```
sampaccoud, line 1234: But these were already our urls 🤔

sampaccoud (approved):

PR #316, opened 2022-10-10
🐛(eucalyptus/3/wb) avoid build error when running npm install

sampaccoud:
## Purpose

When running npm install on `edx-ui-toolkit`, npm tries to run a command:
`/bin/usr/git ls-remote -h -t git://github.com/bessdsv/bower-installer.git`

It seems that github does not accept this syntax anymore because it still works with `ssh://git@github.com`.

## Proposal

Since it's not trivial to control the git commands that npm issues, we can avoid this problem by installing only the specific packet we need.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,11 +158,13 @@ jobs:
   #
   # Note that the job name should match the EDX_RELEASE value
 
-  # Run jobs for the dogwood.3-fun release
-  dogwood.3-fun:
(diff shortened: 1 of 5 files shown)


lunika:
It seems you missed to run the `./bin/ci update` before pushing

sampaccoud:
> It seems you missed to run `./bin/ci update` before pushing

Done!

lunika (approved):

PR #326, opened 2023-05-17
Install ipython xblock and Release eucalyptus.3-wb-1.13.0

sampaccoud:
## Purpose

We need this xblock to run Jupyter. It is already installed in the `dogwood/3/fun image`.

## Proposal

Added Python dependency: `ipython-xblock==0.2.0`

the build was broken because the `faulthandler` packet is not available anymore for Python 2. Since this packet is only useful to run the tests, we think we can skip its installation in the production image, so I added a patch for this.

### DIFF
--- a/releases/eucalyptus/3/wb/CHANGELOG.md
+++ b/releases/eucalyptus/3/wb/CHANGELOG.md
@@ -9,6 +9,10 @@ release.
 
 ## [Unreleased]
 
+### Added
+
+- Add Python dependency `ipython-xblock==0.2.0`
+
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


releases/eucalyptus/3/wb/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ release.
 
 ## [Unreleased]
 
+## [eucalyptus.3-wb-1.13.0] - 2022-05-17
```
jbpenrath, line 12:
```suggestion
## [eucalyptus.3-wb-1.13.0] - 2023-05-17
```

jbpenrath (approved):
LGTM ✅

PR #329, opened 2023-07-03
📌(dogwood/3/fun) pin urlib3 to ensure python 2 compatibility

sampaccoud:
## Purpose

A new version of urlib3 came out which breaks compatibility with Python2. 

## Proposal

Pin `urlib3` dependency to its version 1.


jbpenrath (approved):
LGTM ✅

PR #330, opened 2023-07-04
💄(dogwood/3/fun) use new FUN logo in menu

sampaccoud:
## Purpose

The new FUN logo was recently added in fun-apps but the setting pointing to it is overriden here so the new logo won't show on the site after deployment in production.

## Proposal

Point setting to the new FUN logo.


jbpenrath (approved):
LGTM ✅

PR #335, opened 2023-10-10
Dogwood.fun.3 upgrade fun apps to v5.16.0

sampaccoud:
### Changed

- Upgrade fun-apps to version 5.16.0 to add a title to the course run synchronization webhook data

### Fixed

- Remove dependency to missing Python package `moto` only used in tests


jbpenrath (approved):
LGTM ✅

PR #339, opened 2024-01-16
Configure dogwood with fun-apps 5.17.0 and oauth2 middleware

lunika:
## Purpose

In the fun-apps 5.17.0 we added a new middleware to fix a bug in the Oauth2 process. We must add it in the dogwood MIDDLEWARE_CLASSES settings.

## Proposal

- [x] add oauth2 middleware in the settings
- [x] bump to version 2.11.0


jbpenrath (approved):
LGTM ✅

